PR #2, opened 2019-04-08
Create necessary folders and copy metadata

sutaakar:
Fixing SpringBoot S2I issues


sutaakar:
@mswiderski Can you please take a look?

mswiderski (approved):

mswiderski:
Thanks @sutaakar

PR #13, opened 2019-05-16
Upgrade of GraalVM to rc16

mswiderski:
@spolti could you please review?


spolti (approved):

mswiderski:
@spolti what would be the way to get it build and pushed to quay.io?

spolti:
@mswiderski the images are under the kiegroup organization.
To push, you can execute make push (I think you have permissions there, if not we can ask Babak or Thommy).
I can do this for you after you merge this changes.

mswiderski:
@spolti if you could do the build and push that would be great, thanks a lot

spolti:
@mswiderski I just pushed the new images.

PR #21, opened 2019-06-17
[KOGITO-2] - Remove cekit2 and use only cekit3

spolti:
A few bugs on cekit that was preventing to build and squash images
are now fixed, we can use only cekit3.


spolti:
@mswiderski can you please review?

mswiderski (commented):
@spolti not that I know much about it so maybe @ricardozanini can give his opinion here as well

mswiderski:
I just triggered Jenkins job again...

ricardozanini:
@mswiderski @spolti looks good to me and makes a lot of sense updating the commands to cekit3.

PR #22, opened 2019-06-21
[JBPM-8549] - Introducing Kogito Kubernetes Client

ricardozanini:
See:
https://issues.jboss.org/browse/JBPM-8549

This PR is meant to add a kubernetes client module that doesn't rely on reflection to parse JSON responses, but at the same time uses Fabric8 Kube client as an alternative to perform authentication and configuration of the client within a OpenShift/Kube cluster.

It's meant to be simple to tackle the Kogito Cloud use cases, like service discovery.

Tested against JDK8, JDK11, GraalVM and native images on docker containers and an OpenShift 3.11 cluster.

Signed-off-by: Ricardo Zanini <zanini@redhat.com>

### DIFF
--- a/kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
+++ b/kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
@@ -1,0 +1,10 @@
+# Kogito Cloud Kubernetes Client
+
+This library is designed to be used with Kogito Cloud module, but also as a standalone library as an alternative Kubernetes client on [native images](https://www.graalvm.org/docs/reference-manual/aot-compilation/).
+
+It's a wrapper around the [Fabric8 Kubernetes Client API](https://github.com/fabric8io/kubernetes-client), with some tweaks:
+
+1. Removes the Kube Config file parsing as an alternative to connect to the cluster because of [its reflection usage](https://github.com/fabric8io/kubernetes-client/issues/1591).
+2. Parse JSON API responses as simple Maps to avoid using reflection to build domain objects during runtime like Jackson does. As an alternative, uses Yasson to parse JSON responses and a simple utility to walk into the resulted map.
+
+It's meant to be simple to support simple use cases.
--- a/kogito-cloud-services/kogito-cloud-kubernetes-client/pom.xml
+++ b/kogito-cloud-services/kogito-cloud-kubernetes-client/pom.xml
@@ -7,9 +7,9 @@
     <artifactId>kogito-cloud-services</artifactId>
     <version>8.0.0-SNAPSHOT</version>
   </parent>
-  <artifactId>kogito-cloud-workitems</artifactId>
-  <name>Kogito Cloud Work Items</name>
-  <description>Kogito Cloud Work Items</description>
+  <artifactId>kogito-cloud-kubernetes-client</artifactId>
+  <name>Kogito Cloud Kubernetes Client</name>
+  <description>Kogito Cloud Kubernetes Client</description>
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -17,14 +17,13 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.kie.kogito</groupId>
-      <artifactId>kogito-api</artifactId>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.kie.kogito</groupId>
-      <artifactId>kogito-cloud-kubernetes-client</artifactId>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
     </dependency>
-
     <dependency>
       <groupId>org.junit.vintage</groupId>
       <artifactId>junit-vintage-engine</artifactId>
@@ -40,19 +39,17 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-server-mock</artifactId>
+      <version>${version.kubernetes-client}</version>
+      <scope>test</scope>
+    </dependency>
 
     <!-- Logging -->
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <!-- Fabric8 client used exclusively for tests purposes only -->
-    <dependency>
-      <groupId>io.fabric8</groupId>
-      <artifactId>kubernetes-server-mock</artifactId>
-      <version>${version.kubernetes-client}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/KogitoKubeConfig.java
+++ b/kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/KogitoKubeConfig.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.cloud.kubernetes.client;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import io.fabric8.kubernetes.client.BaseClient;
+import io.fabric8.kubernetes.client.Config;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.dsl.base.OperationSupport;
+import okhttp3.OkHttpClient;
+
+/**
+ * Wraps the Kubernetes Client setup and configuration, exposing infrastructure components to connect to a Kubernetes cluster reliably. 
+ * Most used when there's a need to customize the Fabric8 Kubernetes Client.
+ */
+public final class KogitoKubeConfig {
+
+    public static final String KNATIVE_ISTIO_NAMESPACE = "istio-system";
+    private static final String NAMESPACE_REPLACE = "@ns@";
+    private static final String KNATIVE_SERVICE_SERVICE_URL = "apis/serving.knative.dev/v1alpha1/namespaces/" + NAMESPACE_REPLACE + "/services";
+    private static final String KNATIVE_ISTIO_GATEWAY_URL = "api/v1/namespaces/" + KNATIVE_ISTIO_NAMESPACE + "/services/istio-ingressgateway";
+
+    private KubernetesClient kubernetesClient;
+
+    public KogitoKubeConfig() {
+        // disable kube config file deserialization due to reflection usage
+        System.setProperty(Config.KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, "false");
+        this.kubernetesClient = new DefaultKubernetesClient();
+    }
+
+    /**
+     * Highly customizable config client. Most used in integration tests. 90% of the time you won't need to use this constructor.
+     * 
+     * @param kubernetesClient
+     */
+    public KogitoKubeConfig(KubernetesClient kubernetesClient) {
+        this.kubernetesClient = kubernetesClient;
+    }
+
+    public OkHttpClient getHttpClient() {
+        return ((BaseClient) this.kubernetesClient).getHttpClient();
+    }
+
+    public URL getMasterUrl() {
+        return this.kubernetesClient.getMasterUrl();
+    }
+
+    public String getKNativeServiceServiceURL(String namespace) throws MalformedURLException {
+        if (namespace == null || namespace.isEmpty()) {
+            throw new IllegalArgumentException("A namespace should be provided when using KNative service operations");
+        }
+        final StringBuilder sb = new StringBuilder(this.getMasterUrl().toString());
+        sb.append(KNATIVE_SERVICE_SERVICE_URL.replaceFirst(NAMESPACE_REPLACE, namespace));
+        return sb.toString();
+    }
+
+    public String getKNativeIstioGatewayURL() throws MalformedURLException {
+        final StringBuilder sb = new StringBuilder(this.getMasterUrl().toString());
+        sb.append(KNATIVE_ISTIO_GATEWAY_URL);
+        return sb.toString();
+    }
+
+    public URL getServiceOperationURL(final String namespace) throws MalformedURLException {
+        final OperationSupport services = ((OperationSupport) this.kubernetesClient.services());
+        return services.getNamespacedUrl(namespace);
+    }
+
+}
--- a/kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/MapWalker.java
+++ b/kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/MapWalker.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.cloud.kubernetes.client.operations;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility to avoid chaining conversion in code
+ */
+@SuppressWarnings("unchecked")
+public final class MapWalker {
+
+    private Object theMap;
+    private boolean safeNull;
+
+    /**
+     * Walker with safe null
+     * @param theMap that we're going to walk into
+     * @param safeNull whether to return a null value at the end of the walk, otherwise if the key isn't expected a NPE or an {@link IllegalArgumentException} might raise.
+     */
+    public MapWalker(Object theMap, boolean safeNull) {
+        this.theMap = theMap;
+        this.safeNull = safeNull;
+    }
+
+    public MapWalker(Object theMap) {
+        this(theMap, false);
+    }
+
+    /**
+     * Return the object as a Map.
+     * @param <K>
+     * @param <V>
+     * @return
+     */
+    public <K, V> Map<K, V> asMap() {
+        if (theMap instanceof Map) {
+            return (Map<K, V>) theMap;
+        }
+        throw new IllegalArgumentException(String.format("The object %s is not a map. Impossible to get", theMap));
+    }
+
+    /**
+     * Return the object as a list
+     * @param <K>
+     * @param <V>
+     * @return
+     */
+    public <K, V> List<Map<K, V>> asList() {
+        if (theMap instanceof List) {
+            return (List<Map<K, V>>) theMap;
+        }
+        throw new IllegalArgumentException(String.format("The object %s is not a list. Impossible to get", theMap));
+    }
+
+    /**
+     * The key value is another map 
+     */
+    public MapWalker mapToMap(String key) {
+        if (theMap instanceof Map) {
+            this.theMap = ((Map<String, ?>) theMap).get(key);
+            if (safeNull && this.theMap == null) {
+                this.theMap = new HashMap<>();
+            }
+            return this;
+        }
+        throw new IllegalArgumentException(String.format("The object %s is not a map. Impossible to walk to the key '%s'", theMap, key));
+    }
+
+    /**
+     * The key value is a list map.
+     * 
+     * @param key
+     * @return
+     */
+    public MapWalker mapToListMap(final String key) {
+        if (theMap instanceof Map) {
+            theMap = (List<Map<?, ?>>) ((Map<?, ?>) theMap).get(key);
+            if (safeNull && this.theMap == null) {
+                this.theMap = new ArrayList<Map<?, ?>>();
+            }
+            return this;
+        }
+        throw new IllegalArgumentException(String.format("The object %s is not a list. Impossible to walk to the key %s", theMap, key));
+    }
+
+    /**
+     * In a list, we take the index that is a map
+     * @param index
+     * @return
+     */
+    public MapWalker listToMap(final int index) {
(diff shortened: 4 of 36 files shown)


kogito-cloud-services/kogito-cloud-kubernetes-client/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-jsonb</artifactId>
```
mswiderski: what's the reason to add quarkus here?

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/OperationsResponseParser.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.kogito.cloud.kubernetes.client.operations;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.json.bind.Jsonb;
+import javax.json.bind.JsonbBuilder;
```
mswiderski: isn't it Kubernetes-client uses something else for Json parsing? What I am saying here is that we will have multiple json libraries on class path...

kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
```diff
@@ -0,0 +1,10 @@
+# Kogito Cloud Kubernetes Client
+
+This library is designed to be used with Kogito Cloud module, but also as a standalone library as an alternative Kubernetes client on [native images](https://www.graalvm.org/docs/reference-manual/aot-compilation/).
+
+It's a wrapper around the [Fabric8 Kubernetes Client API](https://github.com/fabric8io/kubernetes-client), with some tweaks:
```
mswiderski, line 5: there seems to be a supported Kubernetes client coming out with quarkus 0.18.0 release, might be worth to have a look

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/OperationsResponseParser.java
ricardozanini: They use Jackson and it's rely a lot on reflection usage, that's why I'm using the same library Quarkus uses to parse JSON. We're just using the Kube client for authentication purposes to avoid reflection at all.

kogito-cloud-services/kogito-cloud-kubernetes-client/pom.xml
ricardozanini: To use their JSON parser instead of Jackson.

kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
ricardozanini: Yeah, I was involved in the discussions there. What they did was to register every domain object for reflection and removing EC (from Bouncy Castle) at all. Same thing that I did in my demos. 

So we have two options: rely on the extension and keep using the reflection or parse the JSON responses manually. I vote for the first one.

I can replace to use it as soon as they release the new version. Will be easy since this library is just a wrapper and the new DiscoveredServiceWorkItem won't be affected at all once I send the other PR (waiting for this to be merged).

ricardozanini:
@mswiderski the code is now refactored and integrated with workitems. After this PR I'm going to work on a demo of how to use this workitem on OCP and also run more integration tests with KNative, Openshift 4 and Native images.

Once we have Quarkus, I should replace the KubernetesServiceDiscovery for the extension. Though, for SpringBoot we also have to support the same client. I don't want to add the Quarkus dependency to this project based on what we discussed. The bright side of having this agnostic implementation is that we can run anywhere, with any framework.

kogito-cloud-services/kogito-cloud-workitems/pom.xml
```diff
@@ -20,6 +20,10 @@
       <groupId>org.kie.kogito</groupId>
       <artifactId>kogito-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-cloud-kubernetes-client</artifactId>
```
mswiderski, line 25: should we remove the fabric8 Kube client as it is dependency not really in the work items module but the kogito-cloud-kubernetes-client instead?

kogito-cloud-services/pom.xml
```diff
@@ -30,6 +33,11 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-cloud-kubernetes-client</artifactId>
+        <version>${version.org.kie.kogito}</version>
+      </dependency>
```
mswiderski: shouldn't this be declared in kogito-bom?

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/ServiceDiscoveryFactory.java
```diff
@@ -0,0 +1,83 @@
...
+    private boolean istioEnv;
+    private String istioGatewayUrl;
+
+    private ServiceDiscoveryFactory(final KogitoKubeClient kubeClient) {
+        this.kubeClient = kubeClient;
+        this.discoverIstioGatewayUrl();
+    }
+
+    public static ServiceDiscovery build() {
+        return build(new DefaultKogitoKubeClient());
```
mswiderski, line 37: just a question - will this create new client and by that check Istio environment every time? If so do we cache the service discovery somehow ... I believe it should be enough to check it once as if it was not there when the service is deployed then we should not expect it will show up afterwards, or?

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/KubernetesServiceDiscovery.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.kie.kogito.cloud.kubernetes.client.KogitoKubeClient;
+import org.kie.kogito.cloud.kubernetes.client.operations.MapWalker;
+import org.kie.kogito.cloud.workitems.ServiceInfo;
+
+class KubernetesServiceDiscovery extends BaseServiceDiscovery {
```
mswiderski: any reason why this is not a public class?

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/IstioServiceDiscovery.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.kie.kogito.cloud.kubernetes.client.KogitoKubeClient;
+import org.kie.kogito.cloud.kubernetes.client.operations.MapWalker;
+import org.kie.kogito.cloud.workitems.ServiceInfo;
+
+class IstioServiceDiscovery extends BaseServiceDiscovery {
```
mswiderski: same here, why it is not public class?

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import org.kie.kogito.cloud.kubernetes.client.KogitoKubeClient;
+import org.kie.kogito.cloud.workitems.ServiceInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+abstract class BaseServiceDiscovery implements ServiceDiscovery {
```
mswiderski: and one more time... why it's not public?

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/DiscoveredServiceWorkItemHandler.java
```diff
@@ -24,125 +24,78 @@
...
         mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
         mapper.setDateFormat(new StdDateFormat().withColonInTimeZone(true));
-        
+
         http = buildHttpClient();
+
+        if (kubeClient == null) {
+            serviceDiscovery = this.buildServiceDiscovery(new DefaultKogitoKubeClient());
+        } else {
+            serviceDiscovery = this.buildServiceDiscovery(kubeClient);
```
mswiderski, line 71: assuming the work item handler instance is cached service discovery instance is created just once, right?

kogito-cloud-services/kogito-cloud-kubernetes-client/src/test/java/org/kie/kogito/cloud/kubernetes/client/KubeClientConfigTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Test
+    public void whenCreateNewConfigurationIgnoresKubeConfigFile() {
+        final KogitoKubeConfig config = new KogitoKubeConfig();
+        assertThat(config.getHttpClient(), notNullValue());
+        assertThat(config.getMasterUrl().toString(), containsString("kubernetes.default.svc"));
+    }
+
+    @Test
+    public void whenGetFromSystemProps() {
+        System.setProperty("kubernetes.master", "localhost");
```
mswiderski, line 63: might be worth to clear the properties in `@After` method just to be on safe side

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/MapWalker.java
```diff
@@ -0,0 +1,121 @@
...
+@SuppressWarnings("unchecked")
+public final class MapWalker {
+
+    private Object theMap;
+    private boolean safeNull;
+
+    /**
+     * Walker with safe null
+     * @param theMap that we're going to walk into
+     * @param safeNull whether to return a null value at the end of the walk, otherwise if the key isn't expected a NPE or IBE might raise.
```
mswiderski: what's IBE? Or you meant IllegalArgumentException ....

kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+1. Removes the Kube Config file parsing as an alternative to connect to the cluster because of [its reflection usage](https://github.com/fabric8io/kubernetes-client/issues/1591).
```
mswiderski, line 7: so we don't have Kube config based auth but we have service account based, right?

mswiderski (commented):
nice work, added few comments and questions

kogito-cloud-services/kogito-cloud-workitems/pom.xml
ricardozanini: Agreed

kogito-cloud-services/pom.xml
ricardozanini: Sure!! Makes sense!

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/KubernetesServiceDiscovery.java
ricardozanini: Overall, I normally don't expose classes that don't need to. But rethinking, someone could override those implementations for whatever need in the future, makes sense turning them to public

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/DiscoveredServiceWorkItemHandler.java
ricardozanini: Right!

kogito-cloud-services/kogito-cloud-kubernetes-client/src/test/java/org/kie/kogito/cloud/kubernetes/client/KubeClientConfigTest.java
ricardozanini: It's cleaning the properties see:
https://github.com/kiegroup/kogito-cloud/pull/22/files/6e4350ef1ee51c30a4e67a776cca260716a3ccea#diff-cfbfa46e4f584b2e82e3e90fec9ffbecR51

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/operations/MapWalker.java
ricardozanini: YES!

kogito-cloud-services/kogito-cloud-kubernetes-client/README.md
ricardozanini: Correct! Kube config relies on reflection, but this won't be needed since the container will be running inside the cluster. If running outside, one could simply pass `-e KUBERNETES_MASTER` and `-e KUBERNETES_OAUTH_TOKEN` variables to the docker run command to authenticate.

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/ServiceDiscoveryFactory.java
ricardozanini: the service discovery is cached in the work item, so assuming that the work item is cached, the service discovery will be cached as well. I didn't want to cache the service discovery instance inside the factory because each work item instance in the container could choose the service discovery mechanism (Operators, KNative or Kubernetes). At the same time, if we have one work item per instance, the service discovery will be cached as well. 

What do you think?

mswiderski: makes perfect sense

kogito-cloud-services/kogito-cloud-kubernetes-client/src/test/java/org/kie/kogito/cloud/kubernetes/client/KubeClientConfigTest.java
mswiderski: it's actually setting it to empty string rather clearing :)

ricardozanini: hahaha true. I'm going to fix that

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/KogitoKubeConfig.java
```diff
@@ -0,0 +1,82 @@
...
+        return ((BaseClient) this.kubernetesClient).getHttpClient();
+    }
+
+    public URL getMasterUrl() {
+        return this.kubernetesClient.getMasterUrl();
+    }
+
+    public String getKNativeServiceServiceURL(String namespace) throws MalformedURLException {
+        final StringBuilder sb = new StringBuilder(this.getMasterUrl().toString());
+        if (namespace == null) {
```
sutaakar: Is null value a valid option for namespace? In that case Knative service URL will be ...namespaces//services , doesn't look like valid value for me.

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Should implement the query logic
+     * 
+     * @param namespace
+     * @param labelKey
+     * @param labelValue
+     * @return
+     */
+    protected abstract List<Map<String, Object>> query(final String namespace, Map<String, String> labels);
```
sutaakar, line 51: This method should return services based on parameters, right? It is not much clear from the method definition or javadoc.

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
```diff
@@ -0,0 +1,88 @@
...
+        } else if (services.isEmpty()) {
+            LOGGER.warn("Haven't found any endpoint in the namespace {} with labels {}:{}", namespace, labelKey, labelValue);
+            return Optional.empty();
+        }
+
+        return Optional.of(this.buildService(services, labelValue == null || labelValue.isEmpty() ? labelKey : labelValue));
+    }
+
+    public final Optional<ServiceInfo> findEndpoint(String namespace, String service) {
+        return this.findEndpoint(namespace, service, null);
```
sutaakar, line 87: If I understand this correctly the service search is converted into search for service with label name equal to name of the service. Is it so?

ricardozanini: Yes, it's. Was done to keep it compatible with the old code. With the other overload (with labels parameters), a client could search for any label. I'll improve the docs.

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
ricardozanini: You're right! I'll improve the docs

kogito-cloud-services/kogito-cloud-kubernetes-client/src/main/java/org/kie/kogito/cloud/kubernetes/client/KogitoKubeConfig.java
ricardozanini: Good point. Should throw an IllegalArgumentException.

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
sutaakar: I am not sure if the "old code" was actually correct. For example in 7.x templates the services contain labels with same name, only label value is used to distinguish services (example: service: "${APPLICATION_NAME}-kieserver"). But in this endpoint we don't provide any service label value, therefore we wouldn't be able to identify correct service.

ricardozanini: I agree and I have put a lot of thought on this. The problem is that only looking for the label key, we indeed, restrict the search. But I'm not willing to change this right now. I'd rather leaving this change to a new JIRA to reengineer the DiscoveredService, so it can look for a key/value label pair. Also, the OnBoarding example should get reviewed as well. I guess that's too much work for now. The intention of this PR is to keep the same interface, only changing the way how we discover the service. WDYT?

sutaakar: Ok, I see. Fine for me as long as we track the work to be done and plan to implement it in future.

ricardozanini:
Onboarding example working on OpenShift 3.11 querying the standard Service API. The onboarding service image (built with the changes of this PR) can be pulled from: https://quay.io/repository/ricardozanini/onboarding-service

@sutaakar if you'd like to test this PR this is the image. The HR and Payroll can be built as well with this PR: https://github.com/kiegroup/kogito-cloud/pull/23

I have s2i images built with GraalVM 19.x if you prefer.

ricardozanini:
I just create a project to make it easy to test this PR:
https://github.com/ricardozanini/onboarding-example-templates

Creating an environment with Istio + KNative for more tests.

sutaakar:
@ricardozanini Thanks, will take a look on this later today or tomorrow morning

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
ricardozanini: @sutaakar just filed a new JIRA to tackle the discussed improvements: https://issues.jboss.org/browse/KOGITO-59

Feel free to add more requirements that you understand it's needed. :)

mswiderski (approved):

mswiderski:
@sutaakar @ricardozanini is this ready to be merged and does it have any dependency to other PRs? Tomorrow is 0.2.0 release so if we want it to be in then it should be merged today

ricardozanini:
@mswiderski @sutaakar I found a bug when using KNative with OCP 4.x. The KNative Serving API [deprecated](https://github.com/knative/serving/blob/master/docs/spec/spec.md#service) the `domain` property from the `status` model, so we need to change it to `url`. It's pretty simple, but it's late night here and I'll add some tests to have it cover. It was the last test that I can think of to merge this PR for 0.2.0.

Nevertheless, this error also happens in the actual version of the DiscoveredWorkItem. I'll support both KNative versions 0.6.x (with `domain`) and 0.7.x. Tomorrow I should update this PR with this fix and the related image in my Quay repo.

Also, I just updated the test repo with KNative instructions on OCP 4.x:
https://github.com/ricardozanini/onboarding-example-templates

@mswiderski this will work for you? I'd be very happy to have this PR for 0.2.0 :heart:

kogito-cloud-services/kogito-cloud-workitems/src/main/java/org/kie/kogito/cloud/workitems/service/discovery/BaseServiceDiscovery.java
ricardozanini: @sutaakar just filed a new JIRA to tackle the discussed improvements: https://issues.jboss.org/browse/KOGITO-59

Feel free to add more requirements that you understand it's needed. :)

ricardozanini:
@mswiderski as talked earlier, I'll update this PR tomorrow morning (15 hours from now) with this fixes and unit tests to cover it so we can have a nice merge for the 0.2.0.

This PR needs to be merged as well:
https://github.com/kiegroup/kogito-bom/pull/36

ricardozanini:
@mswiderski @sutaakar last update to this PR. Things are running smoothly and now we have unit test coverage for Knative Operators + Istio. 
Tested onboarding sample on OCP 3.11, OCP 4.1, OCP 4.1 + KNative Operators.

If there's nothing more to add, we are ok to merge.

PR #23, opened 2019-07-16
[KOGITO-58] Upgrading GraalVM libraries

ricardozanini:
Please see:
https://issues.jboss.org/browse/KOGITO-58

This PR adds a new module 19.x under GraalVM path. I wonder that keeping a *.x could be simpler in the future to upgrade the 19.
There's a lot of enhancements that worth the upgrade, plus the new Quarkus version doesn't work with the old GraalVM.

@spolti @mswiderski could you guys please take a look?


s2i/kogito-quarkus-centos-s2i-overrides.yaml
```diff
@@ -15,7 +15,7 @@ modules:
   install:
   - name: add-kogito-user
   - name: graalvm
-    version: "1.0.0-rc16"
+    version: "19.1.0"
```
mswiderski: currently quarkus only supports 19.0.2 and not 19.1.0 so I'd stay with 19.0.2 for now

ricardozanini: downgraded!

mswiderski (approved):

mswiderski:
let's wait for @spolti review and merge, thanks @ricardozanini

spolti (approved):

spolti:
@mswiderski I'll update the images tomorrow morning, just need to confirm the tag to use, can I tag as 0.2.0 then delete the previous one?

spolti:
Right, just saw your comment on the Jira.

mswiderski:
yes, 0.2.0 makes sense and then remove the old ones. thanks

PR #27, opened 2019-07-29
[KOGITO-65] Use maven.test.skip to skip the test phase

sutaakar:
maven.test.skip makes sure that test jar is not created - https://maven.apache.org/plugins/maven-jar-plugin/test-jar-mojo.html#skip.
@spolti Can you please take a look?

Signed-off-by: Karel Suta <ksuta@redhat.com>


ricardozanini:
Hi @sutaakar! The build was generating the test-jar file even with `skipTests=true`? That's why you added this property?

sutaakar:
@ricardozanini Yes, the skipTests property skips test execution, but not the test jar generation.

ricardozanini:
Good to know. So this property skips tests and the jar generation? Having said that, LGTM.

ricardozanini (approved):

PR #30, opened 2019-08-16
[KOGITO-162] - Migrate the Kogito base images from centos to ubi8

spolti:


spolti:
@mswiderski SHould the GraalVM be updated to 19.1.1?

spolti:
Just tested https://github.com/kiegroup/kogito-examples/tree/0.2.0/drools-quarkus-example on openshift with the new images, works as expected.

ricardozanini:
@spolti we have a JIRA to upgrade to 19.1.1. But once we upgrade the Kogito runtimes to a compatible quarkus version, which is 0.20:

https://issues.jboss.org/browse/KOGITO-98

Feel free to assign this JIRA to you and work on it. Many thanks!

ricardozanini (changes requested):
~The ImageStream also need to be updated?~

Just found the file, lgtm.

ricardozanini (approved):

mswiderski:
Have you tested onboarding example? Reason why i as is that is used https and by that some native library from GraalVM so would be good to check that UBI 8 has all libraries needed to make the https based connections. 

When it comes to graalvm version - I will try to do upgrade of quarkus version soon and then we align graalvm with it. Will let you know

ricardozanini:
@mswiderski thanks for noticing the TLS thing. This is really a PITA, but thankfully they made some changes on 19.1.1/Quarkus 0.20 to avoid the issues we're having.

spolti:
@mswiderski not yet, thanks for remind me, will test it before remove the wip label.
@ricardozanini cool, I can take deal with the graalvm update. 

Thanks.

s2i/modules/add-kogito-user/module.yaml
```diff
@@ -1,8 +1,14 @@
 schema_version: 1
+version: "0.3.0"
+
```
ricardozanini: Can we remove this new line?

s2i/modules/add-kogito-user/module.yaml
```diff
@@ -1,8 +1,14 @@
...
 execute:
 - script: add-user
 
 envs:
 - name: "KOGITO_HOME"
   value: "/home/kogito"
+- name: "USER"
+  value: "kogito"
+
+
```
ricardozanini: double new line

ricardozanini (changes requested):
Micro things. But I guess we should coordinate this merge with https://issues.jboss.org/browse/KOGITO-98

spolti:
@mswiderski @ricardozanini just tried the onboarding example, work as expected.

mswiderski (approved):
looks good to me, if we merge the upgrade of quarkus (https://github.com/kiegroup/kogito-bom/pull/48) then we should upgrade graalvm here as well.

ricardozanini:
@mswiderski we already upgraded to GraalVM 19.1.x and will merge this soon to run e2e tests with https://github.com/kiegroup/kogito-bom/pull/48. If you decide to not merge for 0.3.0, we can rollback this upgrade.
Sorry, I didn't see your comment before.

spolti:
This PR is ready.

PR #32, opened 2019-08-21
[KOGITO-262] - missing version on install-rpm-depencies module

spolti:


ricardozanini:
thanks!

PR #34, opened 2019-08-31
[KOGITO-222] - Export Infinispan authentication options

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-222

Related PR:
https://github.com/kiegroup/kogito-runtimes/pull/142

In this PR we introduce a way to have infinispan authentication system properties imported via env vars. These vars will be set by the Kogito Operator.

Working image for tests:

```
quay.io/ricardozanini/kogito-data-index:latest
```

Includes both PRs.

### DIFF
--- a/s2i/modules/kogito-data-index/added/start
+++ b/s2i/modules/kogito-data-index/added/start
@@ -1,3 +1,38 @@
 #!/bin/bash -e
 
-java -Dquarkus.http.host=0.0.0.0 -Djava.library.path=$KOGITO_HOME/bin -jar $KOGITO_HOME/bin/*-runner.jar
+function set_infinispan_props() {
+    local infinispan_props=""
+
+    if [[ "${INFINISPAN_USEAUTH}" == "true" &&  -z "${INFINISPAN_USERNAME}" ]]; then
+        echo "[ERROR] Flag INFINISPAN_USEAUTH set to true, but no username informed. Please use INFINISPAN_USERNAME and INFINISPAN_PASSWORD variables to set the right credentials."
+        exit 1
+    fi
+
+    if [ -z "${INFINISPAN_USERNAME}" ]; then
+        if [ ! -z "${INFINISPAN_USEAUTH}" ]; then
+            INFINISPAN_USEAUTH="false"
+        fi
+    fi
+ 
+    if [ ! -z "${INFINISPAN_USERNAME}" ]; then infinispan_props=$(echo "${infinispan_props} -Dinfinispan_username=${INFINISPAN_USERNAME}"); INFINISPAN_USEAUTH="true"; fi
+    if [ ! -z "${INFINISPAN_PASSWORD}" ]; then infinispan_props=$(echo "${infinispan_props} -Dinfinispan_password=${INFINISPAN_PASSWORD}"); fi
+    if [ ! -z "${INFINISPAN_USEAUTH}" ]; then infinispan_props=$(echo "${infinispan_props} -Dinfinispan_useauth=${INFINISPAN_USEAUTH}"); fi
+    if [ ! -z "${INFINISPAN_AUTHREALM}" ]; then infinispan_props=$(echo "${infinispan_props} -Dinfinispan_authrealm=${INFINISPAN_AUTHREALM}"); fi
+    if [ ! -z "${INFINISPAN_SASLMECHANISM}" ]; then infinispan_props=$(echo "${infinispan_props} -Dinfinispan_saslmechanism=${INFINISPAN_SASLMECHANISM}"); fi
+
+    echo $infinispan_props
+}
+
+function start() {
+    local infinispan_props=$(set_infinispan_props)
+
+    java -Djava.library.path=$KOGITO_HOME/lib \
+        -Dquarkus.http.host=0.0.0.0 \
+        ${infinispan_props} \
+        ${JAVA_OPTS} \
+        -jar $KOGITO_HOME/bin/*-runner.jar
+}
+
+if [ -z "${TEST}" ]; then
+    start
+fi
--- a/s2i/modules/kogito-data-index/tests/bats/start.bats
+++ b/s2i/modules/kogito-data-index/tests/bats/start.bats
@@ -1,0 +1,94 @@
+#!/usr/bin/env bats
+
+export TEST=true
+
+# import
+load $BATS_TEST_DIRNAME/../../added/start
+
+function clear_vars() {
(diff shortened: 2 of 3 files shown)


ricardozanini:
@mswiderski @spolti could you kindly review this PR?

s2i/modules/kogito-data-index/added/start
```diff
@@ -1,3 +1,11 @@
...
 
-java -Dquarkus.http.host=0.0.0.0 -Djava.library.path=$KOGITO_HOME/bin -jar $KOGITO_HOME/bin/*-runner.jar
\ No newline at end of file
+java -Djava.library.path=$KOGITO_HOME/lib \
+     -Dquarkus.http.host=0.0.0.0 \
+     -Dinfinispan_username=${INFINISPAN_USERNAME} \
+     -Dinfinispan_password=${INFINISPAN_PASSWORD} \
+     -Dinfinispan_useauth=${INFINISPAN_USEAUTH} \
+     -Dinfinispan_authrealm=${INFINISPAN_AUTHREALM} \
+     -Dinfinispan_saslmechanism=${INFINISPAN_SASLMECHANISM} \
```
spolti: is there any side effect if some of then are not set?
or, some of these properties have default values?
or, wouldn't be better set these props only if the credentials are set?

spolti (changes requested):

s2i/modules/kogito-data-index/added/start
ricardozanini: There's no side effect. The infinispan cluster just warns that those properties are empty.

ricardozanini: And the client behaves like it should (e.g. with their default values)

But you made a good point, if those env vars is not set, I'll not set the java property. I don't know about future implementations

mswiderski: +1 to avoid setting them if they are not available, having warnings in the logs is nothing good to start with

ricardozanini:
@spolti could you kindly review the changes?

s2i/modules/kogito-data-index/added/start
```diff
@@ -1,3 +1,27 @@
...
+    local infinispan_props=$(set_infinispan_props)
+
+    java -Djava.library.path=$KOGITO_HOME/lib \
+        -Dquarkus.http.host=0.0.0.0 \
+        ${infinispan_props} \
+        ${JAVA_OPTS} \
+        -jar $KOGITO_HOME/bin/*-runner.jar
+}
+
+if [ -z "${TEST}" ]; then
```
spolti, line 36: its not needed.

s2i/modules/kogito-data-index/tests/bats/start.bats
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bats
+
+# it's a test environment
+export TEST="true"
```
spolti: not needed.

s2i/modules/kogito-data-index/tests/bats/start.bats
```diff
@@ -0,0 +1,49 @@
...
+
+# import
+source $BATS_TEST_DIRNAME/../../added/start
```
spolti: you can use load instead source.

s2i/modules/kogito-data-index/tests/bats/start.bats
```diff
@@ -0,0 +1,49 @@
...
+    unset INFINISPAN_USERNAME
+    unset INFINISPAN_PASSWORD
+    unset INFINISPAN_AUTHREALM
+    unset INFINISPAN_SASLMECHANISM
+}
+
+@test "check if infinispan properties is blank" {
+    clear_vars
+    local expected=""
+    local result=$(set_infinispan_props)
```
spolti: maybe update to the following would be better:

run set_infinispan_props

and grab the output with
${output} or ${lines[@]}

s2i/modules/kogito-data-index/tests/bats/start.bats
```diff
@@ -0,0 +1,49 @@
...
+    
+    echo "Result is ${result} and expected is ${expected}" >&2
+    [ "${expected}" = "${result}" ]
+}
+
+
+@test "check if infinispan auth is false" {
+    clear_vars
+    export INFINISPAN_USEAUTH="false"
```
spolti, line 28: IMHO it should be automatically set if the required envs (user and pwd I suppose) are not set.

s2i/modules/kogito-data-index/added/start
```diff
@@ -1,3 +1,27 @@
 #!/bin/bash -e
 
-java -Dquarkus.http.host=0.0.0.0 -Djava.library.path=$KOGITO_HOME/bin -jar $KOGITO_HOME/bin/*-runner.jar
\ No newline at end of file
+function set_infinispan_props() {
```
spolti, line 3: there are no required properties to configure the auth?
in this way, any property will be set.

i.e. you can set useauth to true and not provide credentials.

spolti (changes requested):

s2i/modules/kogito-data-index/tests/bats/start.bats
ricardozanini: don't need. this one is managed by infinispan if the variables are not set

s2i/modules/kogito-data-index/added/start
ricardozanini: there's no required properties, plus all of this will be defined, managed and configured by the operator. We won't enforce any properties in the image to not have to duplicate rules here and there.

s2i/modules/kogito-data-index/tests/bats/start.bats
ricardozanini: gosh, completely forgot. have been a long time since my last bash. hehehe

s2i/modules/kogito-data-index/added/start
ricardozanini: The `load` function is calling the start function nevertheless. :( 

So I guess we should keep this. If you have another workaround, let me know please.

ricardozanini:
@spolti I guess I covered the most topics we've talked. At least the basic use case is covered and the image could react against common/minor mistakes.

But at the end, this image will be used with the Operator and there those vars will be enforced.

Thanks for all the tips :heart:

spolti (approved):

ricardozanini:
@mswiderski could you take a look?

mswiderski (approved):

PR #35, opened 2019-09-02
[KOGITO-239] - Error after building springboot application with lates…

spolti:
…t ubi8 images

### DIFF
--- a/s2i/modules/kogito-springboot-ubi8-s2i/configure
+++ b/s2i/modules/kogito-springboot-ubi8-s2i/configure
@@ -8,5 +8,5 @@ cp -rv ${SCRIPT_DIR}/s2i/bin /usr/local/s2i
 mkdir $KOGITO_HOME/bin
 
 chmod 755 /usr/local/s2i/*
-chown -R 1001:1001 ${KOGITO_HOME}
+chown -R 1001:0 ${KOGITO_HOME}
 chmod -R ug+rwX ${KOGITO_HOME}
--- a/s2i/modules/kogito-springboot-ubi8-s2i/s2i/bin/usage
+++ b/s2i/modules/kogito-springboot-ubi8-s2i/s2i/bin/usage
@@ -9,11 +9,11 @@ To use it, install S2I: https://github.com/openshift/source-to-image
 
 This image is intended to build application binaries only that are later on transfered to runtime image.
 
-s2i build -c <source code path/URL> quay.io/kiegroup/kogito-springboot-ubi8-s2i <application image name> --runtime-image <image name> --runtime-artifact /home/kogito/bin
+s2i build -c <source code path/URL> quay.io/kiegroup/kogito-springboot-ubi8-s2i <application image name> --runtime-image <image name>
 
 Sample invocation:
 
-s2i build -c /Development/project/ quay.io/kiegroup/kogito-springboot-ubi8-s2i application --runtime-image kogito-springboot-ubi8 --runtime-artifact /home/kogito/bin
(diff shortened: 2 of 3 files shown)


s2i/kogito-imagestream.yaml
```diff
@@ -15,18 +15,18 @@ items:
       openshift.io/provider-display-name: Kie Group.
   spec:
     tags:
-    - name: '0.3.0'
+    - name: '0.3.0-1'
```
sutaakar: is it needed to change imagestream tag name too?

ricardozanini: For this to work now. We have a couple options though:

1. Create this new tag
2. Replace the old 0.3.0
3. Wait until 0.4.0 to have this working on quay

I vote for adding a new tag.

ricardozanini:
@spolti have you tested with the `s2i` CLI as well?

I mean, have removing the `runtime-artifact` from the command could bring any issues on the local build?

spolti:
@ricardozanini yes, both working.

s2i/kogito-imagestream.yaml
spolti: that's a good question, quay.io keeps the two tags distinct, suggestions on how can we improve the tags for small fixes?

@ricardozanini, @mswiderski  ^

sutaakar: If you change an imagestream tag here, AFAIK you need to update Kogito operator too, to point to this new tag, right?
https://github.com/kiegroup/kogito-cloud-operator/blob/master/pkg/controller/kogitoapp/builder/build_config.go#L11

spolti: depends on how operator is getting the image, if it is using latest, there shouldn't be a problem, if it is using the specific tag, yes, then we need to decide how we'll tag new images.

spolti: Then yes, what do you guys suggest?

update the tag to 0.3.0-1 or use latest on operator?

sutaakar: Would current 0.3.0 tag resolve to 0.3.0-1 image when it will be uploaded to quay? Or will it be still pointing to old 0.3.0 images? If it would redirect to 0.3.0-1 then I would keep the version as is.

sutaakar: Or can you remove 0.3.0 tag from old image and put it on new image?

spolti: > Would current 0.3.0 tag resolve to 0.3.0-1 image when it will be uploaded to quay? Or will it be still pointing to old 0.3.0 images? If it would redirect to 0.3.0-1 then I would keep the version as is.

This is basically the same thing than override the tag, probably the best will be still on 0.3.0?

sutaakar: yes

spolti: Alright, pr update, images being updated.

sutaakar:
Looks good to me (not tested by myself). In case of free time I can try to build and run images tomorrow, but if you want to merge it sooner feel free to do so.

spolti:
I am working on behave tests so we can simulate the s2i build.
For openshift we can, maybe, add a pipeline to do some tests on ocp.

mswiderski (approved):

s2i/kogito-imagestream.yaml
ricardozanini: The Operator has now a parameter to change the image name. Will use 0.3.0 only if not receive the new parameter.

ricardozanini (approved):

sutaakar (approved):

PR #37, opened 2019-09-09
[KOGITO-240] - include s2i and some behave tests for the Kogito images

spolti:


sutaakar (approved):
Nice work

s2i/tests/features/kogito-quarkus-index.feature
```diff
@@ -0,0 +1,52 @@
+@quay.io/kiegroup/kogito-data-index
+
+Feature: Kogito-quarkus-ubi8 feature.
```
ricardozanini, line 3: Could you please add a new Scenario with those values for env vars? Might be an error in the image that I'd like to check. Plus these values are closer to a real use case.

```
INFINISPAN_CREDENTIAL_SECRET=infinispan-credentials
INFINISPAN_AUTHREALM=default
INFINISPAN_SASLMECHANISM=PLAIN
QUARKUS_INFINISPAN_CLIENT_SERVER_LIST=server:11222
MP_MESSAGING_INCOMING_KOGITO_PROCESSINSTANCES_EVENTS_BOOTSTRAP_SERVERS=mycluster:9092
KOGITO_PROTOBUF_FOLDER=/home/kogito
KOGITO_PROTOBUF_WATCH=true
INFINISPAN_USEAUTH=true
```

ricardozanini (changes requested):
Could you kindly add this new scenario?

s2i/tests/features/kogito-quarkus-index.feature
spolti: added.

ricardozanini (approved):
LGTM

PR #41, opened 2019-09-17
[KOGITO-301] - Force Kogito runtime images to start on port 8080

spolti:


ricardozanini (approved):
lgtm

My only concern is about that fixed quarkus version in the test-app pom file. Won't that will matter a lot since this is just an app to check the correct exposed port.

mswiderski:
@spolti would it be possible to git clone the repo instead of copying the entire project here?

ricardozanini:
@mswiderski git clone could jeopardize the test if the repo move or someone made a great change on it. But it's debatable to clone it every time and override the `application.properties` file to inject our little "bug" on it to have this test to verify it correctly.

spolti:
This will require changes on CeKit side since we can't tell to cekit to clone before, do some changes and then start the build.

ricardozanini:
@mswiderski this would impact other components to build the image, so cloning the test app is not feasible now.

ricardozanini (approved):

spolti:
@mswiderski can we merge this one?

mswiderski:
> @mswiderski git clone could jeopardize the test if the repo move or someone made a great change on it. But it's debatable to clone it every time and override the application.properties file to inject our little "bug" on it to have this test to verify it correctly.

@ricardozanini it should be taken from Kogito-examples so it I must work anyway and by that we verify that it works based on all the changes in that repo. otherwise we can end up with outdated example here as most likely people will forget about it

> @mswiderski this would impact other components to build the image, so cloning the test app is not feasible now.

can we revisit that? I don't want to duplicate examples that might be easily forgotten ... that's my main point

> @mswiderski can we merge this one?

if we can't avoid duplication and this brings value then I think we should merge but I still would prefer to revisit it and remove the duplicated example

ricardozanini:
@mswiderski your concern is super valid. I also have this appeal. Though, I understand how difficult is to have a feature on cekit since can impact other projects.

What you think about merging this one and opening a JIRA to keep track of this change to add the `git clone` feature to the cekit behave tests?

@spolti ^^^

mswiderski:
Fair enough, feel free to merge and then let’s see if we can have git clone there

spolti:
I totally agree with that cloning the project is better. However it will requires changes on cekit.
I opened this jira to track it: https://issues.jboss.org/browse/KOGITO-312

PR #42, opened 2019-09-19
[KOGITO-303] - Incremental builds in Kogito Runtime images not working

spolti:

### DIFF
--- a/s2i/Makefile
+++ b/s2i/Makefile
@@ -58,20 +58,20 @@ _push:
 .PHONY: push-staging
 push-staging: build _push-staging
 _push-staging:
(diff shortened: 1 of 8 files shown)


mswiderski (approved):

ricardozanini (approved):
I'm just missing a test case for the incremental feature. If would be possible to add, I'd say that would be nice to have it. :)

spolti:
I'll investigate it tomorrow morning.

spolti:
Tests added.
CeKit don't erase the s2i generated images, with that said, we can perform 2 s2i builds in two different test scenario.
The first one will do a full build (full by downloading all dependencies) then save the .m2 for later usage.
The second test, will use the same image then the previously one, and as the both tests does have the incremental tag set, the .m2 will be restored.

ricardozanini (approved):

PR #55, opened 2019-10-23
[KOGITO-419] - Limiting jvm xmx for native in 80% of container memory…

ricardozanini:
… limit

See:
https://issues.jboss.org/browse/KOGITO-419

The rationale behind this change is to try to limit the builder JVM to use only 80% of the container's memory to its heap size, the rest will be handled by the system. Without this change, builds will fail unless it's given an absurd amount of memory.

This first screenshot illustrates the a build before the change:

https://pasteboard.co/IDk0xVn.png

The build fails because the 10Gi of memory was added to the heap directly, with no room for other tasks within the build.

The second screenshot we can see that the build behaves like it should:

https://pasteboard.co/IDk0Ps8.png

The remaining 20% is there to give space to other process to handle the build.

Still, if the memory is set to a lower value (like 4GB), the container will be killed since won't have enough memory for the native-image component itself.

### DIFF
--- a/s2i/modules/graalvm/19.x/added/memory-limit.sh
+++ b/s2i/modules/graalvm/19.x/added/memory-limit.sh
@@ -12,13 +12,18 @@
 # its value must be in binary bytes.
 
 function configure() {
-    # native builds requires at least 1024m (1073741824=1024MB - 1 MB = 2^20 B in base 2)
     # does not accepts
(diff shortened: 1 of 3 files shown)


mswiderski (commented):
looks good to me though I am not the best person to approve it so leave it to the experts

spolti (approved):

PR #58, opened 2019-10-29
[KOGITO-382] - Removing snapshot builds from maven archetypes

ricardozanini:
See:
https://issues.jboss.org/browse/KOGITO-382

Just removed the SNAPSHOT from the archetype generation, introduced the `KOGITO_VERSION` env var to the module and add a new test to verify the maven archetype generation for the `dmn-quarkus-example` example.

### DIFF
--- a/s2i/modules/kogito-quarkus-ubi8-s2i/module.yaml
+++ b/s2i/modules/kogito-quarkus-ubi8-s2i/module.yaml
@@ -3,3 +3,7 @@ version: "0.5.0"
 name: kogito-quarkus-ubi8-s2i
 execute:
   - script: configure
+
+envs:
+- name: "KOGITO_VERSION"
+  value: "0.5.0"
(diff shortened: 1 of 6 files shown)


sutaakar (approved):
Looks good to me.

s2i/modules/kogito-quarkus-ubi8-s2i/module.yaml
```diff
@@ -3,3 +3,7 @@ version: "0.5.0"
 name: kogito-quarkus-ubi8-s2i
 execute:
   - script: configure
+
+envs:
+- name: "KOGITO_VERSION"
+  value: "0.5.0"
```
spolti, line 9: 0.5.1 is already available on maven, shouldn't we use the latest?

spolti (changes requested):

s2i/modules/kogito-quarkus-ubi8-s2i/module.yaml
ricardozanini: Yes, when we bump them all. We have another JIRA to bump the images:
https://issues.jboss.org/browse/KOGITO-521

spolti (approved):

PR #63, opened 2019-11-06
[KOGITO-506] - Incremental builds are not working on OpenShift

spolti:

### DIFF
--- a/s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
+++ b/s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
@@ -1,5 +1,4 @@
 @quay.io/kiegroup/kogito-quarkus-ubi8-s2i
-
 Feature: kogito-quarkus-ubi8-s2i image tests
 
   Scenario: Verify if the s2i build is finished as expected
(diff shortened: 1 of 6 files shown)


ricardozanini:
Looks good, but Jenkins is failing :(

spolti:
the build is failing at microdnf step.
this flag "--setopt=tsflags=nodocs" is causing the issue.

Ongoing discussion on CeKit channel.

ricardozanini:
One last question, I suppose you tested this fix with the operator in OCP?

spolti:
no, the build on my machine was working because the ubi-8 image was outdated. As it is failing now I am evaluating our options.
It will not be a trivial task since the Jenkins use a shared cekit node to build the Kogito images.

ricardozanini:
Oh, understood. So let's wait for a workaround for this issue with Cekit then, because I'd like to see this working on an actual cluster.
Do you think this bug will impact our release next week? :disappointed:

ricardozanini (approved):

spolti:
no, the ubi8/ubi-minimal:8.0 is enough so far, it still graded as A.

s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
```diff
@@ -65,12 +64,17 @@ Feature: kogito-quarkus-ubi8-s2i image tests
 
   Scenario: Perform a incremental s2i build
     Given s2i build https://github.com/kiegroup/kogito-examples.git from drools-quarkus-example with env and incremental using 0.5.1
-    Then check that page is served
```
sutaakar, line 68: @spolti  Why was this check removed?

spolti: I don't recall exactly but probably because there are a few tests doing the same check and the purpose of these two tests are only test incremental builds.

@ricardozanini do you remember if there is another reason behind this removal.

ricardozanini: You've already answered. This check is duplicated and we're removing native tests from the images to speed up the CI and local test process.

sutaakar: ok

PR #75, opened 2019-12-12
[KOGITO-587] - Introducing persistence and k8s module to register pro…

ricardozanini:
…to files to configMaps

See:
https://issues.redhat.com/browse/KOGITO-587

--------8<--------
**TODO**

- [x] - Write behave tests
- [x] - Implement Kubernetes client
- [x] - Implement preHook script that will write to protobuf `configMaps`

--------8<--------

@spolti could you take a look? This is already working for s2i images. I'll test on runtime images and write some behave tests tomorrow, but would like to hear from you if this change follows the best practices that you introduced in the refactoring. 

Here's the resulting directory after a successful build:

```bash
$ ls -la /home/kogito/data/proto/
total 16
drwxrwxr-x. 1 kogito root   4096 Dec 12 22:30 .
drwxrwxr-x. 1 kogito root   4096 Dec 12 22:04 ..
-rw-r--r--. 1 kogito kogito  870 Dec 12 22:30 demo.orders.proto
-rw-r--r--. 1 kogito kogito  796 Dec 12 22:30 persons.proto
``` 

And here's the `launch` dir:

```bash
$ ls -la /home/kogito/launch/
total 28
drwxrwxr-x. 2 kogito root 4096 Dec 12 22:04 .
drwxrwx---. 1 kogito root 4096 Dec 12 22:04 ..
-rw-rw-r--. 1 kogito root 2363 Dec 12 22:04 configure-maven.sh
-rw-rw-r--. 1 kogito root 4245 Dec 12 22:03 configure.sh
-rwxrwxr-x. 1 kogito root 1744 Dec 12 22:04 kogito-persistence.sh
-rw-rw-r--. 1 kogito root 1586 Dec 12 22:03 memory-limit.sh
```


ricardozanini:
It seems that the file was not copied correctly: 
``` 
ERROR:root:I1212 19:03:17.675842   25391 postexecutorstep.go:357] /usr/local/s2i/s2i-core: line 4: /home/kogito/launch/kogito-persistence.sh: No such file or directory
```

Will take a look at the Jenkins error tomorrow and hopefully fixed it once I write the behave tests.

s2i/modules/kogito-persistence/added/kogito-persistence.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# move_persistence_files moves persistence files from $KOGITO_HOME/bin to the final directory
+# where those files will be handled by the runtime image. 
+# TODO: remove this function when s2i build move the KOGITO_HOME instead bin directory between images in chained builds
+function move_persistence_files() {
+    echo "---> [persistence] Moving persistence files to final directory"
+    if ls $KOGITO_HOME/bin/*.proto &>/dev/null 
+    then
+        # copy to the final dir, so we keep bin clean
+        cp -v $KOGITO_HOME/bin/*.proto $KOGITO_HOME/data/proto/
```
spolti: mv instead cp?
or do we need the .proto files under bin directory as well?

s2i/modules/kogito-kubernetes-client/module.yaml
```diff
@@ -0,0 +1,6 @@
+schema_version: 1
+name: org.kie.kogito.kubernetes.client
```
spolti, line 2: what is the purpose of this module?
didn't find its configure script and there is no usage of it.

spolti (changes requested):
looks great, just a few comments.

s2i/modules/kogito-kubernetes-client/module.yaml
ricardozanini: I'm working on it

s2i/modules/kogito-persistence/added/kogito-persistence.sh
ricardozanini: I tried to use `mv`, but got permissions issue (looks like the runtime user doesn't have permissions in the dir). It's working like the `image_metadata.json` file.

spolti: ok, in the future we re-work this as we will move the image_metadata away from kogito_home/bin.

ricardozanini:
@spolti could you please take a final look? =)

s2i/modules/kogito-data-index/configure
```diff
@@ -7,6 +7,8 @@ ADDED_DIR=${SCRIPT_DIR}/added
 
 mv ${SOURCES_DIR}/data-index-service-*-runner.jar ${KOGITO_HOME}/bin/
 mv ${ADDED_DIR}/* ${KOGITO_HOME}/bin/
+mkdir -p ${KOGITO_HOME}/data/protobufs/
+chmod +w ${KOGITO_HOME}/data/protobufs/
```
spolti: not needed since we are adding rw to user an group below.

s2i/modules/kogito-jq/module.yaml
```diff
@@ -0,0 +1,17 @@
...
+version: "1.5"
+
+artifacts:
+    - name: jq-release.key
+      url: https://raw.githubusercontent.com/stedolan/jq/master/sig/jq-release.key
+      md5: a7adb1c4acf92078d730cde007b7e5d8
+    - name: jq-linux64.asc
+      url: https://raw.githubusercontent.com/stedolan/jq/master/sig/v1.5/jq-linux64.asc
+      md5: 8113abb78708f005f4b915feda6c612a
+    - name: jq-linux64
```
spolti: from the e-mail thread it is not clear yet if we can or not use this.

s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
```diff
@@ -43,6 +43,17 @@ Feature: kogito-quarkus-ubi8-s2i image tests
...
 
+  Scenario: Verify if the s2i build is finished as expected performing a non native build with persistence enabled
+    Given s2i build https://github.com/kiegroup/kogito-examples.git from jbpm-quarkus-example using 0.6.1 and runtime-image quay.io/kiegroup/kogito-quarkus-jvm-ubi8:latest
+      | variable          | value |
+      | NATIVE            | false   |
+      | MAVEN_ARGS_APPEND | -Ppersistence |
+    Then file /home/kogito/bin/jbpm-quarkus-example-0.6.1-runner.jar should exist
+    And s2i build log should contain '/home/kogito/bin/demo.orders.proto' -> '/home/kogito/data/protobufs/demo.orders.proto'
+    And s2i build log should contain '/home/kogito/bin/persons.proto' -> '/home/kogito/data/protobufs/persons.proto'
+    And s2i build log should contain ---> [persistence] generating md5 for persistence files
```
spolti, line 54: wouldn't be good also compare if the generated md5 is the same than the expected?

s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats
```diff
@@ -0,0 +1,58 @@
...
+teardown() {
+    rm -rf ${KOGITO_HOME}
+    rm -rf /tmp/src
+    rm -rf $KOGITO_HOME/bin/*
+}
+
+@test "Call post hook script (sanity check)" {
+    MOCK_RESPONSE="config_map.json" # we have annotations, but no files in the file system
+    
+    local expected=$(cat $BATS_TEST_DIRNAME/expected/patch_empty_data.json)
```
spolti: where is the MOCK_RESPONSE and expected envs used?

s2i/modules/kogito-persistence/tests/bats/kogito-persistence.bats
```diff
@@ -0,0 +1,174 @@
...
+    run update_configmap
+    
+    echo "result= ${lines[@]}"
+    [ "$status" -eq 0 ]
+    [ "${lines[1]}" = "INFO ---> [persistence] About to patch configMap exampleapp-cm" ]
+    [ "${lines[2]}" = "Body: ${expected}" ]
+}
+
+@test "Patch with an empty annotations configmap with files in disk" {
+    MOCK_RESPONSE="config_map_empty_annotations.json" # we have empty annotations and files in disk
```
spolti, line 160: what is the purpose of this env?

s2i/modules/kogito-persistence/added/post-hook-persistence.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Post Hook script that should be called by Kogito Service Kubernetes Pod upon initialization
+# Updates the Kogito Service protobuf configMap with the persistence files located at ${KOGITO_HOME}/data/protobufs
+
+source $KOGITO_HOME/launch/kogito-persistence.sh
+
+update_configmap
```
spolti, line 8: If it needs to be run only on the startup why not add it on the script that runs the image?
like this: https://github.com/kiegroup/kogito-cloud/blob/master/s2i/modules/kogito-jobs-service/added/kogito-app-launch.sh#L15

spolti (changes requested):

s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
ricardozanini: We have to read the generated files since there's no messages in the logs. There's no way to do that via behave tests.

s2i/modules/kogito-persistence/added/post-hook-persistence.sh
ricardozanini: This is the script that will be called by [kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#define-poststart-and-prestop-handlers), should be run by the launch scripts.

s2i/modules/kogito-jq/module.yaml
ricardozanini: Yes, this is here for testing purposes, the final one should be the RH's and I'll update the module as soon as I have the reply from the maintainer.

s2i/modules/kogito-persistence/tests/bats/kogito-persistence.bats
ricardozanini: See: https://github.com/kiegroup/kogito-cloud/pull/75/files?file-filters%5B%5D=.bats&file-filters%5B%5D=.feature&file-filters%5B%5D=.json&file-filters%5B%5D=.proto&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension#diff-94a714f7d87d7c00e980e84751e815dbR31

s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats
ricardozanini: See: https://github.com/kiegroup/kogito-cloud/pull/75/files?file-filters%5B%5D=.bats&file-filters%5B%5D=.feature&file-filters%5B%5D=.json&file-filters%5B%5D=.proto&file-filters%5B%5D=.yaml&file-filters%5B%5D=No+extension#diff-94a714f7d87d7c00e980e84751e815dbR31

ricardozanini: It's mocking the k8s API response.

s2i/tests/features/kogito-quarkus-ubi8-s2i.feature
spolti: what about:
```And run sh -c 'cat /home/kogito/bin/proto' in container and immediately check its output for md5```
?

File: s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats (hunk 6 above)
Comment (spolti) [reply s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats]: So I guess you forget to add a check to see if the expected is the same than the result of run $BATS_TEST_DIRNAME/../../added/post-hook-persistence.sh ?

File: s2i/tests/features/kogito-quarkus-ubi8-s2i.feature (hunk 5 above)
Comment (ricardozanini) [reply s2i/tests/features/kogito-quarkus-ubi8-s2i.feature:54]: Bah, I forgot that we can run commands inside the container. Shame on me, it's xmas time. I'm on lazy mode.

File: s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats (hunk 6 above)
Comment (ricardozanini) [reply s2i/modules/kogito-persistence/tests/bats/post-hook-persistence.bats]: Actually we don't need those mocks for this check since the script will import inside an environment we don't have access to in this bats.

I'll remove them completely from this test. This one is just to verify if there's anything wrong with the script file, like a lame typo. The full test is covered in `kogito-persistence.bats`

**Comment by ricardozanini** [comment]
Added proposed changes and rebased

File: s2i/modules/kogito-jq/module.yaml (hunk 4 above)
Comment (spolti) [reply s2i/modules/kogito-jq/module.yaml]: looks like we still waiting for this, right?
Apparently we don't have a final response yet.

Comment (ricardozanini) [reply s2i/modules/kogito-jq/module.yaml]: No, I forwarded the email to the legal department, but no reply so far. I’m on PTO and should get back on Tuesday to follow this up. If you have any news meanwhile, let me know.

**Comment by ricardozanini** [comment]
@spolti changed to EPEL instead of relying from Github binaries. During productization, we can override the `kogito-jq` project to install a module like `org.kie.kogito.rhel8`. Then the `jq` binary will come from the product part.

@mwringe @mswiderski FYI

**Comment by ricardozanini** [comment]
Bah, this error is occurring randomly, and the cause might be this:
https://github.com/rpm-software-management/microdnf/issues/27
https://github.com/cekit/cekit/pull/560

~@spolti do you know which version are we using for this Jenkins?~

EDIT: `[37m2020-01-08 18:10:29,511 cekit        DEBUG    Running version 3.2.0[0m`

Is it possible to update it?

**Comment by ricardozanini** [comment]
@spolti by our internal discussion, we can go ahead with the EPEL solution for jq.

File: s2i/tests/features/kogito-quarkus-ubi8-s2i.feature [hunk 9]
```diff
@@ -43,6 +43,20 @@ Feature: kogito-quarkus-ubi8-s2i image tests
       | expected_phrase | Mario is older than Mark |
     And file /home/kogito/bin/drools-quarkus-example-0.6.1-runner.jar should exist
 
+  @wip
```
spolti: maybe a leftover from tests?

spolti (changes requested):
just a minor observation.

spolti (approved):
Great work, thanks!
approved ;)

ricardozanini:
Let's wait for Jenkins to finish their job :p

PR #80, opened 2019-12-17
[KOGITO-682] - Add integration with a CI to run the Kgito Images bats…

spolti:
… tests

### DIFF
--- a/s2i/modules/kogito-s2i-core/tests/bats/s2i-core.bats
+++ b/s2i/modules/kogito-s2i-core/tests/bats/s2i-core.bats
@@ -7,12 +7,14 @@ setup() {
     export KOGITO_HOME=/tmp/kogito
     export HOME=$KOGITO_HOME
     mkdir -p ${KOGITO_HOME}
+    mkdir target
 }
 
 teardown() {
     rm -rf ${KOGITO_HOME}
     rm -rf /tmp/.s2i
     rm -rf /tmp/src
+    rm -rf target
     rm -rf $KOGITO_HOME/bin/*
 }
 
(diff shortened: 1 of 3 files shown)


s2i/modules/kogito-s2i-core/tests/bats/s2i-core.bats
```diff
@@ -22,15 +24,27 @@ teardown() {
...
+    echo "result= ${sorted[@]}"
 
     [ "$status" -eq 0 ]
-    [ "${lines[0]}" = "Expanding artifacts from incremental build..." ]
-    [ "${lines[1]}" = "./" ]
-    [ "${lines[2]}" = "./file3" ]
-    [ "${lines[3]}" = "./file2" ]
-    [ "${lines[4]}" = "./file1" ]
-    [ "${lines[5]}" = "./file" ]
+    if [ -z "${GITHUB_ACTIONS}" ]; then
```
ricardozanini, line 31: Why? GH Actions mess with the log output?

ricardozanini (changes requested):
Just one question regarding this condition of the bats running on GH

s2i/modules/kogito-s2i-core/tests/bats/s2i-core.bats
spolti: GH actions 'sort' behaves somehow differently than running it locally. the lines even sorted has a different order, not sure why.
I don't know why GH have this behaviour but this is the way I found to keep the same test working local and here.

ricardozanini (approved):

sutaakar (approved):

PR #82, opened 2019-12-19
[KOGITO-705] - Refactor data-index image in the same way kogito-jobs-…

spolti:
…service image is structured


spolti:
images tagged as rc2.
@sutaakar do you mind to run some tests with the new images?
behave and bats and a quick test are passing, just to double check.

ricardozanini (approved):

PR #83, opened 2019-12-26
[KOGITO-511] - Kogito Data Index image - allow quarkus.http.port to b…

spolti:
…e defined via env variable


ricardozanini (approved):
/LGTM

On Operator side, we might need to:
1. Read this property value from `.Spec.Envs` on `KogitoDataIndex` CR and set the probes and ports accordingly; **or**
2. Remove this property from `.Spe.Envs` to respect the meta data set in the image

sutaakar (approved):

spolti:
I would vote on 1.
Will open a PR to follow this up.

PR #87, opened 2020-01-20
[KOGITO-920] - Release Kogito Images 0.7.0

spolti:

### DIFF
--- a/s2i/image.yaml
+++ b/s2i/image.yaml
@@ -1,7 +1,7 @@
 schema_version: 1
 
 name: "kogito-image-real-name-on-overrides-file"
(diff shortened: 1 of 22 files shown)


sutaakar (approved):

ricardozanini (approved):
Looks good, but it seems that we have a problem in CI.

spolti:
totally not related with the problem I spotted on my machine.

>   features/image/kogito-springboot-ubi8-s2i.feature:5  Verify if the s2i build is finished as expected
>   features/image/kogito-springboot-ubi8-s2i.feature:19  Verify if the s2i build is finished as expected using multi-module build

investigating.

PR #89, opened 2020-02-07
[KOGITO-1023] consolidate kogito-bom and kogito-runtimes repos

mariofusco:


pom.xml
```diff
@@ -51,14 +51,14 @@
       <dependency>
```
MarianMacik, line 53: Still inheriting from kogito-bom-parent.

MarianMacik (changes requested):

MarianMacik:
Also please delete `kogito-bom` stage from `Jenkinsfile`:
https://github.com/kiegroup/kogito-cloud/blob/6ad6afe989c8d4bddd2c1016007ed63851442c3b/Jenkinsfile#L21-L30

PR #100, opened 2020-03-06
[KOGITO-1377] Enable batch mode for S2I build

sutaakar:
Signed-off-by: Karel Suta <ksuta@redhat.com>

### DIFF
--- a/s2i/modules/kogito-s2i-core/added/s2i-core
+++ b/s2i/modules/kogito-s2i-core/added/s2i-core
@@ -78,6 +78,10 @@ function build_kogito_app() {
         nativeBuild=""
     fi
 
+    if [ "${SCRIPT_DEBUG}" != "true" ]; then
+        export MAVEN_ARGS_APPEND="${MAVEN_ARGS_APPEND} --batch-mode"
+    fi
+
     if [ -f "pom.xml" ]; then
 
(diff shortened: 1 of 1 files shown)


sutaakar:
There are no bats test attached as this change affects env variable MAVEN_ARGS_APPEND in function build_kogito_app. The build_kogito_app function is triggered in bats test using 'run' function, which prevents retrieving of env variables (or at least I didn't find a way to retrieve it).

ricardozanini (approved):

s2i/modules/kogito-s2i-core/added/s2i-core
```diff
@@ -80,6 +80,10 @@ function build_kogito_app() {
 
     if [ -f "pom.xml" ]; then
 
+        if [ "${SCRIPT_DEBUG}" != "true" ]; then
```
spolti: can't we move it outside the if so we can reuse it on both kind of builds?

spolti (changes requested):

s2i/modules/kogito-s2i-core/added/s2i-core
sutaakar: Done

spolti (approved):

PR #110, opened 2020-03-16
[KOGITO-754] - Create image for Management Console

spolti:
https://issues.redhat.com/browse/KOGITO-754

### DIFF
--- a/s2i/kogito-imagestream.yaml
+++ b/s2i/kogito-imagestream.yaml
@@ -125,7 +125,7 @@ items:
             description: Runtime image for the Kogito Data Index Service
             iconClass: icon-jbpm
             tags: kogito,data-index
-            supports: infinispan, persitence
+            supports: persistence backed by Infinispan server
             version: '0.9.0-rc1'
           referencePolicy:
             type: Local
@@ -137,13 +137,13 @@ items:
     metadata:
       name: kogito-jobs-service
       annotations:
-        openshift.io/display-name: Runtime image for the Kogito Jobs Service - Tech preview
+        openshift.io/display-name: Runtime image for the Kogito Jobs Service
         openshift.io/provider-display-name: Kie Group.
     spec:
       tags:
         - name: '0.9.0-rc1'
           annotations:
-            description: Runtime image for the Kogito Jobs Service - Tech preview
+            description: Runtime image for the Kogito Jobs Service
             iconClass: icon-jbpm
             tags: kogito,jobs-service
             supports: out-of-box process timers
@@ -153,3 +153,25 @@ items:
           from:
             kind: DockerImage
             name: quay.io/kiegroup/kogito-jobs-service:0.9.0-rc1
+  - kind: ImageStream
+    apiVersion: v1
+    metadata:
+      name: kogito-management-console
+      annotations:
+        openshift.io/display-name: Runtime image for the Kogito Management Console
+        openshift.io/provider-display-name: Kie Group.
+    spec:
+      tags:
+        - name: '0.9.0-rc1'
+          annotations:
+            description: Runtime image for the Kogito Management Console
+            iconClass: icon-jbpm
+            tags: kogito,management-console
+            supports: business process management
+            version: '0.9.0-rc1'
+          referencePolicy:
+            type: Local
+          from:
+            kind: DockerImage
+            name: quay.io/kiegroup/kogito-management-console:0.9.0-rc1
+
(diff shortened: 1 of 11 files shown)


s2i/modules/kogito-management-console/added/launch/kogito-management-console.sh
```diff
@@ -0,0 +1,29 @@
...
+    unset KOGITO_DATAINDEX_URL
+}
+
+function configure() {
+    configure_data_index_url
+}
+
+# Exit codes:
+#   10 - invalid url
+function configure_data_index_url {
```
ricardozanini, line 15: @cristianonicolai aren't we going to have HTTP and WS protocols as well? For Runtimes we have:

```
KOGITO_DATAINDEX_HTTP_URL
KOGITO_DATAINDEX_WS_URL
```

@spolti sorry, I know you've asked before, but I just realized that we actually have those both envs.

cristianonicolai: @ricardozanini I can adjust that if you want to use the same naming, but for now we only need HTTP one.

ricardozanini: @cristianonicolai I’d appreciate that, since we are going to use the same bits to inject those variables to the Kogito Services.

spolti: IMHO we could use the same env name and set it using http or ws protocol...
When the console accepts WS we can only add support to this protocol in the iamge.

ricardozanini: The problem is that the target service needs both URLs set. They need to prepare their side to receive this one URL and use it for both protocols. But it's not what they are doing right now, IIRC.

cristianonicolai: @spolti @ricardozanini sorry for the confusion, but the management console is already using `KOGITO_DATAINDEX_HTTP_URL` as the env parameter.

spolti: Ok, for now, will use only KOGITO_DATAINDEX_HTTP_URL.
When it is done on management console side I add the WS one.

Can you guys please review?
Need to merge it before rename this repo.

ricardozanini (approved):

s2i/tests/features/kogito-management-console.feature
```diff
@@ -0,0 +1,32 @@
...
+    When container is started with command bash
+    Then run sh -c 'ls /home/kogito/bin/kogito-management-console-runner.jar' in container and immediately check its output for /home/kogito/bin/kogito-management-console-runner.jar
+
+  Scenario: Verify if the debug is correctly enabled and test default http port
+    When container is started with env
+      | variable     | value |
+      | SCRIPT_DEBUG | true  |
+    Then container log should contain + exec java -XshowSettings:properties -Dkogito.dataindex.http.url=http://localhost:8180 -Dquarkus.http.host=0.0.0.0 -jar /home/kogito/bin/kogito-management-console-runner.jar
+
+  Scenario: Verify if the debug is correctly enabled and and set data-index url
```
MarianMacik: Just a typo `and and`

sutaakar (approved):
Looks good to me.

s2i/kogito-imagestream.yaml
```diff
@@ -137,13 +137,13 @@ items:
...
       annotations:
-        openshift.io/display-name: Runtime image for the Kogito Jobs Service - Tech preview
+        openshift.io/display-name: Runtime image for the Kogito Jobs Service
         openshift.io/provider-display-name: Kie Group.
     spec:
       tags:
         - name: '0.9.0-rc1'
           annotations:
-            description: Runtime image for the Kogito Jobs Service - Tech preview
+            description: Runtime image for the Kogito Jobs Service
```
MarianMacik, line 146: Just noticed a typo on kogito-data-index image stream as well. There is `supports: infinispan, persitence` on the line 128. Maybe would be good to update it?

MarianMacik (approved):
Looks good to me, just 2 typos.

s2i/kogito-imagestream.yaml
spolti: do you mean the space?

spolti: updated.

MarianMacik: I mean `persitence` (missing s).

PR #127, opened 2020-04-07
[0.9.x][KOGITO-1673] - Images should deploy version without micro version in tag

spolti:
See: https://issues.redhat.com/browse/KOGITO-1673
cherry picked from: https://github.com/kiegroup/kogito-images/pull/124


radtriste:
can you add the branch into the title and also correct the title with the one from the issue ?

add also the other PR from which it has been cherry-picked

spolti:
I would prefer use labels for it, for this one we can do this way, will setup new labes later for it.

radtriste:
for the version as usually master and the branch (0.9.x) will have the same name, it is a way to distinguish in emails/notifications ...

cherry-pick can stay into the labels.

also, please update the title with the one from issue: `Images should deploy version without micro version in tag`

spolti:
right, forget the tittle :)

radtriste (approved):

PR #128, opened 2020-04-07
[master][KOGITO-1636] - Native binary is not recognized in binary bui…

spolti:
…ld when target folder is uploaded
See: https://issues.redhat.com/browse/KOGITO-1636

Bats tests on this PR: https://github.com/kiegroup/kogito-images/pull/120/

### DIFF
--- a/modules/kogito-s2i-core/added/s2i-core
+++ b/modules/kogito-s2i-core/added/s2i-core
@@ -58,9 +58,16 @@ function runtime_assemble() {
             echo "-----> Entire target dir uploaded"
             cd "target"    
         fi
-        if ls | grep -E '*-runner.jar' 1> /dev/null 2>&1; then
-            echo "-----> Found jar file, not native build"
-            NATIVE="FALSE"
+
+        local result=$(file -b --mime-type *-runner | grep x-executable)
+        if [ "$result" == "application/x-executable" ]; then
+            echo "-----> Found binary file, native build."
+            NATIVE="TRUE"
+        else
+            if ls | grep -E '*-runner.jar' 1> /dev/null 2>&1;then
+                echo "-----> Found jar file, not native build."
+                NATIVE="FALSE"
+            fi
         fi
         echo "-----> Cleaning up unneeded jar files"
         rm -rfv *-tests.jar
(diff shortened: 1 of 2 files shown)


spolti:
tag 0.10.0-rc1 available.

sutaakar (approved):

radtriste:
@spolti with which artifacts did you build the 0.10.0-rc1 ?
I cannot see anything in git history about that rc1 ...

asking because jobs-service 0.10.0-rc1 is still failing because of oidc...

modules/kogito-s2i-core/added/s2i-core
```diff
@@ -58,9 +58,16 @@ function runtime_assemble() {
...
+        local result=$(file -b --mime-type *-runner | grep x-executable)
+        if [ "$result" == "application/x-executable" ]; then
+            echo "-----> Found binary file, native build."
+            NATIVE="TRUE"
+        else
+            if ls | grep -E '*-runner.jar' 1> /dev/null 2>&1;then
+                echo "-----> Found jar file, not native build."
+                NATIVE="FALSE"
+            fi
         fi
```
radtriste, line 71: should we fail if none of them happen ? (no jar or no native runner)

sutaakar: Well, not necessarily.
There are 3 different possibilities expected:
- non native Quarkus binary (*runner.jar) - covered here
- native Quarkus binary (*runner) - covered here
- Springboot (or any other, plain *.jar file) - no need to cover here as it doesn't have native option

IMHO this part of code decides whether uploaded content is related to native build or not, relevant just for Quarkus.

radtriste: oki. forgot about Spring boot.

radtriste (approved):

sutaakar (changes requested):
~I tested this change with 0.10.0-rc1 image.
The build passed, however when result image was started it crash with
`Error: Unable to access jarfile /home/kogito/bin/*runner.jar`~

I forgot to set native build flag in CR.
The result image runs correctly.

sutaakar (approved):

spolti:
@radtriste I didn't update any snapshot version of jobs-service, data-index or mgmt-console, I can do a separete PR for it and update only these three images, does it sounds good?

radtriste:
yeah would be great to have intermediate 0.10.0-rcX images with latest snapshot as long as we don't have the pipeline ;)

ricardozanini (approved):

PR #145, opened 2020-04-21
[0.9.x][KOGITO-1948] Updated to 0.9.2-rc1

radtriste:
https://issues.redhat.com/browse/KOGITO-1948

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -1,7 +1,7 @@
 schema_version: 1
 
 name: "kogito-image-real-name-on-overrides-file"
-version: "0.9.1"
+version: "0.9.2-rc1"
 # until this issue is not fixed use 8.0 tag.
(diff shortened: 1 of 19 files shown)


radtriste:
will updated with new snapshots version once https://github.com/kiegroup/kogito-apps/pull/193 is merged and deployed into snapshot repository

sutaakar (approved):

ricardozanini:
Are we going to have a `0.9.2`?

radtriste:
not sure, but we should be ready in case :)
I think this should be automatic once we release the previous one

spolti:
@radtriste it should be.For next releases let's do this way.
same PR but different commits.

spolti (approved):

PR #151, opened 2020-04-27
KOGITO-1964 Kogito Images nightly builds

Kaitou786:
This pipeline will build and push the images to quay
Between 12:00AM-3:00AM  UTC  if there is a push on that day

### DIFF
--- a/Jenkinsfile-Nightly
+++ b/Jenkinsfile-Nightly
@@ -1,0 +1,105 @@
+//Checks master branch for updates every day
+//If found triggers the pipeline, 
+//The pipeline is set to run between 12:00AM - 3:00AM UTC
+
+pipeline{
+    agent { label 'jenkins-slave'}
+     triggers{
+        pollSCM(
+            scmpoll_spec: '@midnight', 
+            ignorePostCommitHooks: true
+        )
+    }
+    stages{
+        stage('Initialization'){
+            steps{
+                sh "docker rm -f \$(docker ps -a -q) || docker rmi -f \$(docker images -q) || date"
+            }
+        }
+        stage('Prepare offline kogito-examples'){
+            steps{
+                sh "make clone-repos"
+            }
+        }
+        stage('Build and test kogito-quarkus-ubi8 image'){
+            steps{
+                sh "make kogito-quarkus-ubi8"
+            }
+        }
+        stage('Build and test kogito-quarkus-jvm-ubi8 image'){
+            steps{
+                sh "make kogito-quarkus-jvm-ubi8"
+            }
+        }
+        stage('Build and test kogito-quarkus-ubi8-s2i image'){
+            steps{
+                sh "make kogito-quarkus-ubi8-s2i"
+            }
+        }
+        stage('Build and test kogito-springboot-ubi8 image'){
+            steps{
+                sh "make kogito-springboot-ubi8"
+            }
+        }
+        stage('Build and test kogito-springboot-ubi8-s2i image '){
+            steps{
+                sh "make kogito-springboot-ubi8-s2i"
+            }
+        }
+        stage('Build and test kogito-data-index image '){
+            steps{
+                sh "make kogito-data-index"
+            }
+        }
+        stage('Build and test kogito-jobs-service image '){
+            steps{
+                sh "make kogito-jobs-service"
+            }
+        }
+        stage('Build and test kogito-management-console image '){
+            steps{
+                sh "make kogito-management-console"
+            }
+        }
+        stage('Tagging'){
+            steps{
+                sh """
+                docker tag quay.io/kiegroup/kogito-quarkus-ubi8:latest                    quay.io/kiegroup/kogito-quarkus-ubi8-nightly:\$(echo \${GIT_COMMIT} | cut -c1-7)
+	        docker tag quay.io/kiegroup/kogito-quarkus-jvm-ubi8:latest                quay.io/kiegroup/kogito-quarkus-jvm-ubi8-nightly:\$(echo \${GIT_COMMIT} | cut -c1-7)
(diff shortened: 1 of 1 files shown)


radtriste:
If we plan to have to have kogito-images nightlies, then we should have a way to retrieve latest build artifacts for data-index, jobs-service and management-console ;)

Kaitou786:
I think the retrieval of latest build artifacts for data-index, jobs-service and management-console will be a part of https://github.com/spolti/kie-cloud-tools/tree/master/cekit-cache-server , there is a similar process for rhpam and rhdm

radtriste:
ok is see.
But that pipeline is not useful if we do not update the jar artifacts references IMHO ...

Kaitou786:
IMO I think for now we can let the pipeline build the images from the master as we are already updating those artifacts manually, it'll be good to have the images available on quay for every day when there is some change on the repo

radtriste:
so the build would be run only if changes in repo ?
can you add that into the jenkinsfile ?

Kaitou786:
On midnight the jenkins first check if the commit hash of the master is updated or not, if updated it'll trigger the pipeline. I think the jenkinsfile is meant to run on every PR, and does not push the images on quay

Kaitou786:
Which part are you referring to add in Jenkinsfile?

radtriste:
I was referring to `Jenkinsfile-Nightly`

Kaitou786:
The trigger is done using the jenkins configuration, it does not need to be part of Jenkinsfile-Nightly

radtriste:
I mean it could be set into the jenkinsfile itself: https://www.jenkins.io/doc/book/pipeline/syntax/#triggers (see `pollSCM`)
so we know how it works without having to check in Jenkins configuration

Kaitou786:
I don't think it would be a good idea, by this we can define the cron job, but AFAIK not able to identify if there is a change in the branch, we are not using the periodical build in jenkins config

radtriste:
but the `pollSCM` is able to know if there was a change:

```
pollSCM
Accepts a cron-style string to define a regular interval at which Jenkins should check for new source changes. If new changes exist, the Pipeline will be re-triggered. For example: triggers { pollSCM('H */4 * * 1-5') }
```

radtriste:
@msmagnanijr wdyt ?

Kaitou786:
Yes, we are using the pollSCM in the jenkins configuration, I'll need to run some tests and see if this works

Kaitou786:
I think using it in the jenkins configuration will be better, we are thinking of implementing the shared library in our jenkins enviornment, it'll be a bit difficult if the trigger is a part of the Jenkinsfile-Nightly itself

radtriste:
Ok then.
Can you at least leave a comment at the beginning of the file to explain when this one is launched ?

Kaitou786:
Okay, I'll add that

ricardozanini (approved):

spolti:
I do prefer have everything on the Jenkins file, it will make our lifes eaiser during migrations.
@Kaitou786 @msmagnanijr can everything be configured on the Jenkinsfile?

msmagnanijr:
> I do prefer have everything on the Jenkins file, it will make our lifes eaiser during migrations.
> @Kaitou786 @msmagnanijr can everything be configured on the Jenkinsfile?

t is possible to use "cron" on Jenkisfile but the functionality to check for new updates is done by the plugin that is in Jenkins.

Kaitou786:
I tested adding the triggers in the Jenkinsfile, it'll do the configuration at the Jenkins side automatically, so I guess adding it in the Jenkinsfile will be better for migrations. On the first run of the pipeline, the triggers are automatically set in Jenkins configuration

radtriste:
@msmagnanijr yeah that is the point of `pollSCM`. In fact, almost all the configuration you set in Jenkins configuration can be set in Jenkinsfile. Like @Kaitou786 wrote, it then configures the job on the first run.

@Kaitou786 looks good to me. Do you think we can have something similar for the kogito-cloud-operator-nightly pipeline ? I mean triggers into the file directly ?

radtriste (approved):

Kaitou786:
> @msmagnanijr yeah that is the point of `pollSCM`. In fact, almost all the configuration you set in Jenkins configuration can be set in Jenkinsfile. Like @Kaitou786 wrote, it then configures the job on the first run.
> 
> @Kaitou786 looks good to me. Do you think we can have something similar for the kogito-cloud-operator-nightly pipeline ? I mean triggers into the file directly ?

Yes, I think we'll just need to add the triggers section in that pipeline as well :)

msmagnanijr (approved):

spolti:
this pipeline seems not to be right, the latest commit on master was yesterday and there is 2 tags in less than 24 hours.

https://quay.io/repository/kiegroup/kogito-jobs-service-nightly?tab=tags

plus, none of the commit ids in the tags refer to a real commit.

Can you please verify it?

Kaitou786:
> this pipeline seems not to be right, the latest commit on master was yesterday and there is 2 tags in less than 24 hours.
> 
> https://quay.io/repository/kiegroup/kogito-jobs-service-nightly?tab=tags
> 
> plus, none of the commit ids in the tags refer to a real commit.
> 
> Can you please verify it?

Actually, currently I ran this pipeline manually yesterday to test if everything is okay or not, and again today after adding the triggers section. Currently as this Jenkinsfile is not on kogito-images repo https://jenkins-kogito-tools.apps.kogito.automation.rhmw.io/job/Kogito/job/kogito-images-nightly/configure I configured it to read it from my forked repo, after this is merged, it'll show the commit code of master. I'll change the configuation to read the jenkinsfile from the kogito-images repo. For now we can delete those tags, but we'll have to run this pipeline manually for the first time for configuration to be done on the jenkins master after it is merged over here. After that it'll only run on the day when there is some change on the master branch of kogito-images repo

spolti (approved):

PR #152, opened 2020-05-05
[KOGITO-1214] - Do not use echo statement on the images to print log …

vaibhavjainwiz:
…messages

### DIFF
--- a/modules/kogito-data-index/tests/bats/kogito-data-index.bats
+++ b/modules/kogito-data-index/tests/bats/kogito-data-index.bats
@@ -1,118 +1,23 @@
 #!/usr/bin/env bats
 
-export KOGITO_HOME=/tmp/kogito
-export HOME=$KOGITO_HOME
-mkdir -p ${KOGITO_HOME}/launch
-cp $BATS_TEST_DIRNAME/../../../kogito-infinispan-properties/added/kogito-infinispan-properties.sh ${KOGITO_HOME}/launch/
-
 # imports
-load ${KOGITO_HOME}/launch/kogito-infinispan-properties.sh
 load $BATS_TEST_DIRNAME/../../added/launch/kogito-data-index.sh
 
-teardown() {
-    rm -rf ${KOGITO_HOME}
+setup() {
+    prepareEnv
 }
 
-function clear_vars() {
-    unset INFINISPAN_USEAUTH
-    unset INFINISPAN_USERNAME
-    unset INFINISPAN_PASSWORD
-    unset INFINISPAN_AUTHREALM
-    unset INFINISPAN_SASLMECHANISM
-}
-
-@test "check if infinispan properties is blank" {
-    clear_vars
-    local expected=" -Dquarkus.infinispan-client.use-auth=false"
-    configure_infinispan_props
-    echo "Result is ${INFINISPAN_PROPERTIES} and expected is ${expected}" >&2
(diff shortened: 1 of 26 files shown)


modules/kogito-infinispan-properties/tests/bats/kogito-infinispan-properties.bats
```diff
@@ -1,20 +1,18 @@
...
+export KOGITO_HOME=/tmp/kogito
+mkdir -p ${KOGITO_HOME}/launch
+cp $BATS_TEST_DIRNAME/../../../kogito-logging/added/logging.sh ${KOGITO_HOME}/launch/
 
 # import
 load $BATS_TEST_DIRNAME/../../added/kogito-infinispan-properties.sh
 
-function clear_vars() {
-    unset INFINISPAN_USEAUTH
-    unset INFINISPAN_USERNAME
```
spolti, line 10: why are you removing it?

modules/kogito-data-index/tests/bats/kogito-data-index.bats
```diff
@@ -1,118 +1,23 @@
...
-
 # imports
-load ${KOGITO_HOME}/launch/kogito-infinispan-properties.sh
 load $BATS_TEST_DIRNAME/../../added/launch/kogito-data-index.sh
 
-teardown() {
-    rm -rf ${KOGITO_HOME}
-}
-
-function clear_vars() {
```
spolti, line 16: why did you remove these tests?

modules/kogito-jobs-service/tests/bats/kogito-jobs-service.bats
```diff
@@ -53,19 +48,13 @@ teardown() {
     [ "${KOGITO_JOBS_PROPS}" = "${expected}" ]
 }
 
-
 @test "check if the persistence is correctly configured with auth" {
     export ENABLE_PERSISTENCE="true"
     export INFINISPAN_CLIENT_SERVER_LIST="localhost:11222"
-    export INFINISPAN_USEAUTH="true"
```
spolti, line 60: why removing it?

spolti (changes requested):

modules/kogito-infinispan-properties/tests/bats/kogito-infinispan-properties.bats
vaibhavjainwiz: I replaced it with setup() function which is provided by BATS. Setup function automatically calls before executing every test.

function setup() {
    prepareEnv
    unset INFINISPAN_PROPERTIES
}

modules/kogito-data-index/tests/bats/kogito-data-index.bats
vaibhavjainwiz: I had removed some bats test from kogito-data-index.bats file because these are redundant. kogito-data-index file contains some tests of kogito-infinispan-properties which are already covered in kogito-infinispan-properties.bats file

modules/kogito-jobs-service/tests/bats/kogito-jobs-service.bats
vaibhavjainwiz: I removed it because it tests kogito-inifinispan-properties which are already tested in kogito-infinispan-properties.bats file

vaibhavjainwiz:
incorporate review comments and raised new pull request
https://github.com/kiegroup/kogito-images/pull/153

PR #154, opened 2020-05-11
[KOGITO-2118] - Write script to retrieve latest snapshot urls for services jar artifacts

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-2118

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -7,7 +7,7 @@ Today we have these scripts:
 - [manage-kogito-version.py](manage-kogito-version.py)
 - [push-local-registry.sh](push-local-registry.sh)
 - [push-staging.py](push-staging.py)
-
+- [update-artifacts.py](update-artifacts.py)
 
 ### Managing Kogito images version
 
@@ -85,3 +85,34 @@ This script is called as the last step of the `make push-staging` command define
 It will look for the current RC images available on [quay.io](https://quay.io/organization/kiegroup) to increase the rc tag 
 accordingly then push the new tag so it can be tested by others. 
 
+#### Updating Kogito Images service artifacts.
+
+The update-artifacts script will help in fetching the URL of latest artifacts from the repo
+and update them in the `module.yaml` files of the kogito services
+
+#### Script dependencies
+
+The `update-artifacts.py` has some dependencies that needs to be manually installed:
+
+```bash
+$ pip install -U ruamel.yaml
+$ pip install -U elementpath
+```
+It's usage is pretty simple as well:
+
+```bash
+$ python update-artifacts.py
+```
+
+This script also accepts `--snapshot-version` as argument in specifying the current version of artifacts which need to be fetched
+
+```bash
+$ python update-artifacts.py --snapshot-version='8.0.0-SNAPSHOT'
+```
+if no argument is given, it takes the default value of `8.0.0-SNAPSHOT`
+
+The command will update the needed files with latest snapshot URL:
+
+ - kogito-data-index/module.yaml
+ - kogito-jobs-service/module.yaml
+ - kogito-management-console/module.yaml
--- a/scripts/update-artifacts.py
+++ b/scripts/update-artifacts.py
@@ -1,0 +1,120 @@
+#!/usr/bin/python3
+#Script responsible for fetching the latest artifacts of kogito services and updating their module.yaml files
+#Should be run from root directory of the repository
+#Sample usage:  python3 scripts/update-artifacts.py
+#
+#Dependencies
+# ruamel.yaml
+# elementpath 
+
+
+import xml.etree.ElementTree as ET
+import requests
+import subprocess as sp
+import os
+import argparse
+from ruamel.yaml import YAML
+
+repoURL = "https://repository.jboss.org/org/kie/kogito/"
+
+Modules = {
+    #service-name: module-name(directory in which module's module.yaml file is present)
+    #Note: Service name should be same as given in the repository
+    "data-index-service": "kogito-data-index",
+    "jobs-service": "kogito-jobs-service",
+    "management-console": "kogito-management-console"
+}
+
+
+def getMetadataRoot(repoURL):
+    '''
+    Get the root element from the maven-metadata
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: root object
+    '''
+    metadataURL=repoURL+"maven-metadata.xml"
+    mavenMetadata=requests.get(metadataURL)
+    with open('maven-metadata.xml', 'wb') as f:
+        f.write(mavenMetadata.content)
+    tree = ET.parse('maven-metadata.xml')
+    root=tree.getroot()
+    return root
+
+def getSnapshotVersion(repoURL):
+    '''
+    parse the xml and finds the snapshotVersion
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: snapshotVersion string
+    '''
+    root=getMetadataRoot(repoURL)
+    snapshotVersion=root.find("./versioning/snapshotVersions/snapshotVersion/value").text
+    return snapshotVersion
+
+def getArtifactID(repoURL):
+    '''
+    parse the xml and finds the artifactID
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: artifactID string
+    '''
+    root=getMetadataRoot(repoURL)
+    artifactID=root.find("./artifactId").text
+    return artifactID
+
+def getRunnerURL(repoURL):
+    ''' 
+    Creates the updated URL for runner.jar
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: url string
+    '''
+    artifactId=getArtifactID(repoURL)
+    snapshotVersion=getSnapshotVersion(repoURL)
+    url=repoURL+"{}-{}-runner.jar".format(artifactId,snapshotVersion)
+    return url
+
+def getMD5(repoURL):
+    '''
+    Fetches the md5 code for the latest runner.jar
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: runnerMD5 string
+    '''
+    runnerURL=getRunnerURL(repoURL)
+    runnerMD5URL=runnerURL+".md5"
+    runnerMD5=sp.getoutput("curl -s  {}".format(runnerMD5URL))
+    return runnerMD5
+
+def yaml_loader():
+    '''
+    default yaml Loader
+    :return: yaml object
+    '''
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    yaml.width = 1024
+    yaml.indent(mapping=2, sequence=4, offset=2)
+    return yaml
+
+def update_artifacts(repoURL,modulePath):
+    '''
+    Updates the module.yaml file of services with latest artifacts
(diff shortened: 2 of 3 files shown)


scripts/update-artifacts.py
```diff
@@ -0,0 +1,110 @@
...
+
+def getMetadataRoot(repoURL):
+    '''
+    Get the root element from the maven-metadata
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: root object
+    '''
+    metadataURL=repoURL+"maven-metadata.xml"
+    mavenMetadata=requests.get(metadataURL)
+    open("maven-metadata.xml","wb").write(mavenMetadata.content)
```
r00ta: Hi, what about 
```
with open('maven-metadata.xml', 'wb') as f:
     f.write(mavenMetadata.content)
```
? It would be good to close the file anyway, wdyt?

Kaitou786: Done, thanks :)

sutaakar (approved):
Looks good to me though I have minimal knowledge of Python.

scripts/update-artifacts.py
```diff
@@ -0,0 +1,111 @@
...
+    :parm repoURL: URL of the repo from which artifacts needs to be updated
+    :param modulePath: relative file location of the module.yaml for the kogito service
+    '''
+
+    with open(modulePath) as module:
+        data=yaml_loader().load(module)
+        data['artifacts'][0]['url']=getRunnerURL(repoURL)
+        data['artifacts'][0]['md5']=getMD5(repoURL)
+    with open(modulePath, 'w') as module:
+            yaml_loader().dump(data, module)
```
radtriste: indentation

scripts/update-artifacts.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python3
+#Script responsible for fetching the latest artifacts of kogito services and updating their module.yaml files
```
radtriste, line 2: Can you also add a section in `scripts/README.md` for this script ?

scripts/update-artifacts.py
```diff
@@ -0,0 +1,111 @@
...
+
+def getMetadataRoot(repoURL):
+    '''
+    Get the root element from the maven-metadata
+    :param repoURL: URL of the repo from which artifacts needs to be updated
+    :return: root object
+    '''
+    metadataURL=repoURL+"maven-metadata.xml"
+    mavenMetadata=requests.get(metadataURL)
+    with open('maven-metadata.xml', 'wb') as f:
```
radtriste, line 37: File should be deleted at the end of the script execution

scripts/update-artifacts.py
```diff
@@ -0,0 +1,111 @@
...
+# elementpath 
+
+
+import xml.etree.ElementTree as ET
+import requests
+import subprocess as sp
+from ruamel.yaml import YAML
+
+
+DATA_INDEX_repoURL="https://repository.jboss.org/org/kie/kogito/data-index-service/8.0.0-SNAPSHOT/"
```
radtriste: We may have some nightlies on release branch, which the version `8.0.0-SNAPSHOT` could change.
Can add it as a parameter with `8.0.0-SNAPSHOT` as default ?

scripts/update-artifacts.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python3
+#Script responsible for fetching the latest artifacts of kogito services and updating their module.yaml files
+#Should be run from root directory of the repository
```
radtriste, line 3: Could you add the script execution to the `Jenkinsfile-Nightly` execution as a new stage ?
so we have a nightly pipeline with latest found artifacts ?

radtriste (changes requested):

scripts/README.md
```diff
@@ -85,3 +85,35 @@ This script is called as the last step of the `make push-staging` command define
...
+$ pip install -U elementpath
+```
+It's usage is pretty simple as well:
+
+```bash
+$ python update-artifacts.py
+```
+
+This script also accepts `--snapshot-version` as argument in specifying the current version of the repository
+from which the artifacts needs to be fetched
```
radtriste:
```suggestion
This script also accepts `--snapshot-version` as argument in specifying the current version of artifacts which need to be fetched
```

scripts/README.md
```diff
@@ -85,3 +85,35 @@ This script is called as the last step of the `make push-staging` command define
...
+
+```bash
+$ python update-artifacts.py --snapshot-version='8.0.0-SNAPSHOT'
+```
+if the argument is not passed it takes the default value of `8.0.0-SNAPSHOT`
```
radtriste:
```suggestion
if no argument is given, it takes the default value of `8.0.0-SNAPSHOT`
```

scripts/update-artifacts.py
```diff
@@ -0,0 +1,118 @@
...
+    with open(modulePath, 'w') as module:
+        yaml_loader().dump(data, module)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Update the snapshot url for kogito services.')
+    parser.add_argument('--snapshot-version', dest='snapshotVersion', default='8.0.0-SNAPSHOT', help='Defines the snapshot version of the jboss repository, defaults to 8.0.0-SNAPSHOT')
+    args = parser.parse_args()
+
+    DATA_INDEX_repoURL="https://repository.jboss.org/org/kie/kogito/data-index-service/{}/".format(args.snapshotVersion)
```
radtriste: could we keep the base url as main variables at the beginning of the script and use them here to get correct final url ?

also I think we could factorize a bit to avoid repetition of repository base url

scripts/update-artifacts.py
```diff
@@ -0,0 +1,118 @@
...
+    JOBS_SERVICE_repoURL="https://repository.jboss.org/org/kie/kogito/jobs-service/{}/".format(args.snapshotVersion)
+    MANAGEMENT_CONSOLE_repoURL="https://repository.jboss.org/org/kie/kogito/management-console/{}/".format(args.snapshotVersion)
+
+    DATA_INDEX_modulePath="modules/kogito-data-index/module.yaml"
```
radtriste: Same for module path.

scripts/update-artifacts.py
```diff
@@ -0,0 +1,119 @@
...
+    with open(modulePath, 'w') as module:
+        yaml_loader().dump(data, module)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Update the snapshot url for kogito services.')
+    parser.add_argument('--snapshot-version', dest='snapshotVersion', default='8.0.0-SNAPSHOT', help='Defines the snapshot version of the jboss repository, defaults to 8.0.0-SNAPSHOT')
+    args = parser.parse_args()
+
+    DATA_INDEX_repoURL=repoURL+"data-index-service/{}/".format(args.snapshotVersion)
```
radtriste: We could make this more dynamic by having specific artifact's name and module's name in a struct and iterate on it.
wdyt ?

Kaitou786: Updated the script to be more dynamic by having a struct and iterating over it, if this is okay I'll sqaush the commits :)

scripts/update-artifacts.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+import argparse
+from ruamel.yaml import YAML
+
+repoURL = "https://repository.jboss.org/org/kie/kogito/"
+
+Modules = {
+    #service-name: module-path
+    #Note: Service name should be same as given in the repository
+    "data-index-service": "modules/kogito-data-index/module.yaml",
```
radtriste: I think we could just have `kogito-data-index` as module name and build the module path from there

scripts/update-artifacts.py
```diff
@@ -0,0 +1,127 @@
...
+    '''
+
+    with open(modulePath) as module:
+        data=yaml_loader().load(module)
+        data['artifacts'][0]['url']=getRunnerURL(repoURL)
+        data['artifacts'][0]['md5']=getMD5(repoURL)
+    with open(modulePath, 'w') as module:
+        yaml_loader().dump(data, module)
+
+def updateAllArtifacts(snapshotVersion):
```
radtriste: This could be happening into the main method IMO. no need to have a specific method for it.

radtriste (changes requested):

scripts/update-artifacts.py
Kaitou786: I picked `data-index-service` as it is easier to make URL with this name

spolti:
the script looks good for me, however, how are you planning to create the pull request to update the artifacts?
Does Jenkins have some plugin to do it?

Kaitou786:
> the script looks good for me, however, how are you planning to create the pull request to update the artifacts?
> Does Jenkins have some plugin to do it?

I think the main goal for this script was to update the artifacts in nightly pipelines, if we want to use this for creating PR on a periodic basis, I'll have to check. Do we want to update and send the PR using this script on a nightly basis as well?  @radtriste  wdyt?

scripts/README.md
```diff
@@ -85,3 +85,34 @@ This script is called as the last step of the `make push-staging` command define
...
+
+#### Script dependencies
+
+The `update-artifacts.py` has some dependencies that needs to be manually installed:
+
+```bash
+$ pip install -U ruamel.yaml
+$ pip install -U elementpath
+```
+It's usage is pretty simple as well:
```
Prashansa-K, line 101: Grammatical Error: 
*Its not It's

radtriste:
The main goal was indeed to be able to retrieve latest artifacts and build the images from them for the nightly pipeline.
The creation of PR can be implemented later as part of https://issues.redhat.com/browse/KOGITO-1874. But the PR creation/merge will be implemented into the release pipeline. So I would wait for it and then we should just activate it for nightlies. @spolti are you ok with that ?

radtriste (approved):
lgtm

spolti:
@radtriste we can address it on the other issue.
but the problem of not updating the master branch with the latest is that we always will have to do it manually when building it to test.

The way it works on RHPAM is, when a nightly build is done, a scheduled that runs twice a day (IRRC) will check for it, if a newer build is found than the build on master branch, it will download the artifacts to the cekit cacher, retrieve its md5, and create the pull request to update it.

We could leavarage on this same app done for PAM, but, instead use a scheduler expose a rest endpoint or anything else that can send a payload containing the artifatac's name, checksum and url after the nightly build is successfully done, like the latest step.

spolti (approved):

radtriste:
@spolti This is the whole point of the full nightly pipeline we are putting in place ... The main pipeline will run the runtimes deploy pipeline and then call the images deploy pipeline. And that is where we are calling this script to update the artifacts url/md5.

What I was saying is that we are going to create a possibility to create/merge PRs with the release pipeline. Which means we could also leverage this possibility in the nightly pipeline once implemented. So we may be able later to update the urls/md5 via PR in the nightly pipeline so artifacts are updated for other tests like PR checks.
So the PR creation will come :) and we can have automatic merging in the `images-promote` pipeline so we don't even need human intervention to validate the PR ;)

If you want to implement already as part of this PR, we can, but it is not the main purpose of the issue here. We should implement as part of https://issues.redhat.com/browse/KOGITO-1874

spolti:
I am ok with this approach, about the automatic merge, we need to have at least manual approval :)

PR #157, opened 2020-05-12
[KOGITO-877] - Improve the KOGITO images push-stating script to also …

Kaitou786:
…increase the tag on image-stream
See: https://issues.redhat.com/browse/KOGITO-877

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -8,6 +8,7 @@ Today we have these scripts:
 - [push-local-registry.sh](push-local-registry.sh)
 - [push-staging.py](push-staging.py)
 - [update-artifacts.py](update-artifacts.py)
+- [common.py](common.py)
 
 ### Managing Kogito images version
 
@@ -21,7 +22,7 @@ The `manage-kogito-version.py` has one dependency that needs to be manually inst
 ```bash
 $ pip install -U ruamel.yaml
 ```
-
+This script has a dependency on `common.py`.
 Its usage is pretty simple, only one parameter is accepted:
 
 ```bash
@@ -72,12 +73,20 @@ $ /bin/sh scripts/push-local-registry.sh my_registry_address 0.10 my_namespace
 Staging images are the release candidates which are pushed mainly after big changes that has direct impact on how
 the images will behave and also when new functionality is added.
 
+The script updates the version on:
+
+ - all cekit modules
+ - image.yaml file descriptor
+ - kogito-imagestream.yaml
+ 
+
 #### Script dependencies
 
-The `push-stating.py` has a few dependencies that probably needs to be manually installed:
+The `push-staging.py` has a few dependencies that probably needs to be manually installed:
 
 ```bash
 $ pip install -U docker yaml
+$ pip install -U ruamel.yaml
 ```
 
 This script is called as the last step of the `make push-staging` command defined on the [Makefile](../Makefile).
@@ -116,3 +125,5 @@ The command will update the needed files with latest snapshot URL:
  - kogito-data-index/module.yaml
  - kogito-jobs-service/module.yaml
  - kogito-management-console/module.yaml
+### Common script
+The `common.py` script defines some common functions for the scripts.
--- a/scripts/common.py
+++ b/scripts/common.py
@@ -1,0 +1,112 @@
+#!/usr/bin/python3
+#This script defines some common function that are used by manage-kogito-version.py and push-staging.py script
+
+
+import os
+import re
+
+from ruamel.yaml import YAML
+
+# all kogito-image modules that points to the kogito version.
+MODULES = {"kogito-data-index", "kogito-image-dependencies",
+           "kogito-infinispan-properties", "kogito-jobs-service",
+           "kogito-jq", "kogito-kubernetes-client",
+           "kogito-launch-scripts", "kogito-logging",
+           "kogito-management-console", "kogito-persistence",
+           "kogito-quarkus", "kogito-quarkus-jvm",
+           "kogito-quarkus-s2i", "kogito-s2i-core",
+           "kogito-springboot", "kogito-springboot-s2i",
+           "kogito-system-user"}
+
+# imagestream file that contains all images, this file aldo needs to be updated.
+IMAGE_STREAM = "kogito-imagestream.yaml"
+
+# image.yaml file definition that needs to be updated
+IMAGE = "image.yaml"
+
+# declared envs on modules.yaml that also needs to have its version updated
+ENVS = {"KOGITO_VERSION"}
+
+def yaml_loader():
+    """
+    default yaml Loader
+    :return: yaml object
+    """
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    yaml.width = 1024
+    yaml.indent(mapping=2, sequence=4, offset=2)
+    return yaml
+
+def update_image_version(target_version):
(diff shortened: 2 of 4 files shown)


scripts/README.md
```diff
@@ -4,19 +4,19 @@ On this directory you can find some python scripts used to help with some repeti
 
 Today we have these scripts:
 
-- [manage-kogito-version.py](manage-kogito-version.py)
+- [manage_kogito_version.py](manage_kogito_version.py)
```
radtriste: why changing `-` with `_` ?
all other scripts are named with `-` ...

scripts/push-staging.py
```diff
@@ -125,7 +126,16 @@ def find_current_rc_version():
     if 'QUAY_TOKEN' not in  os.environ:
         print("Env QUAY_TOKEN not found, aborting...")
         os._exit(1)
+
     find_current_rc_version()
+
+    version = CURRENT_IMAGE_VERSION
+    version = version.rstrip(version[-1]) + str(int(version[-1])+1)
```
radtriste: you are also doing the same call on line#57, can you factorize like `getNextVersion(originalVersion)` ?

scripts/push-staging.py
```diff
@@ -125,7 +126,16 @@ def find_current_rc_version():
...
+
+    manage_kogito_version.update_image_version(version)
```
radtriste: instead of importing from another executable script, could we put common methods in a common script ?

radtriste (changes requested):

scripts/README.md
Kaitou786: Python doesn't allow to import modules with `-` in name. So had to change just for importing it.

scripts/push-staging.py
Kaitou786: Okay,I'll create a function for this.

scripts/push-staging.py
Kaitou786: Do you mean to have common functions in a separate script and have `push-staging` and `manage-kogito-version` script import that and run these functions?

radtriste: yep

scripts/README.md
radtriste: the common script should solve this problem

scripts/README.md
```diff
@@ -72,16 +73,27 @@ $ /bin/sh scripts/push-local-registry.sh my_registry_address 0.10 my_namespace
...
+$ pip install -U ruamel.yaml
 ```
 
 This script is called as the last step of the `make push-staging` command defined on the [Makefile](../Makefile).
 
 It will look for the current RC images available on [quay.io](https://quay.io/organization/kiegroup) to increase the rc tag 
 accordingly then push the new tag so it can be tested by others. 
 
+### Common script
+The `common.py` script defines some common functions used by `manage-kogito-version.py` and `push-staging.py`
```
radtriste:
```suggestion
The `common.py` script defines some common functions for the scripts.
```

scripts/README.md
```diff
@@ -72,16 +73,27 @@ $ /bin/sh scripts/push-local-registry.sh my_registry_address 0.10 my_namespace
 Staging images are the release candidates which are pushed mainly after big changes that has direct impact on how
 the images will behave and also when new functionality is added.
 
+The script imports `common.py` for updating the version on:
```
radtriste:
```suggestion
The script updates the version on:
```

scripts/README.md
```diff
@@ -21,7 +22,7 @@ The `manage-kogito-version.py` has one dependency that needs to be manually inst
 ```bash
 $ pip install -U ruamel.yaml
 ```
-
+This script also imports `common.py` for updating version on the respective files
```
radtriste:
```suggestion
This script has a dependency on `common.py`.
```

scripts/common.py
```diff
@@ -0,0 +1,112 @@
...
+#This script defines some common function that are used by manage-kogito-version.py and push-staging.py script
+
+
+import os
+import re
+
+from ruamel.yaml import YAML
+
+# all kogito-image modules that points to the kogito version.
+MODULES = {"kogito-data-index/module.yaml", "kogito-image-dependencies/module.yaml",
```
radtriste: I know you just copy/paste from the other script but could we aboid repeating `module.yaml` here ?
I think we could just have the module name and build the path to `module.yaml` into the function using that array.
wdyt ?

scripts/push-staging.py
```diff
@@ -120,12 +121,25 @@ def find_current_rc_version():
...
     print("Current initial version is %s" % CURRENT_IMAGE_VERSION)
 
+def get_next_rc_version(current_rc_version):
+    '''
+    After finding the current rc tag of the image, adds one to it
+    e.g: 0.10.0-rc1 will returned as 0.10.0-rc2
+    :param current_rc_version: takes the current rc version of the image as input
+    :return: returns the next rc version of the image
+    '''    
+    return (current_rc_version.rstrip(current_rc_version[-1]) + str(int(current_rc_version[-1])+1))
```
radtriste: `-1` could be problematic if we have `rc19` or ?
we would then move to `rc110` or ?

scripts/push-staging.py
```diff
@@ -120,12 +121,25 @@ def find_current_rc_version():
         CURRENT_IMAGE_VERSION = version+'-rc1'
     print("Current initial version is %s" % CURRENT_IMAGE_VERSION)
```
radtriste: could we have a return in this method instead of defining a global variable ?
It took me some time to find where that global var was coming from ...

scripts/push-staging.py
Kaitou786: Do you think that we'll go upto rc19 for any image?

scripts/common.py
Kaitou786: Yeah, make sense I'll update that

scripts/push-staging.py
radtriste: we should make it generic, so handling any number after `rc`

Kaitou786: Okay, I'll update the function

radtriste (approved):

spolti (approved):
Great work Tarun!

spolti:
@Kaitou786  Can you please rebase?

Kaitou786:
@spolti Thanks and done! :)

spolti:
let's wait the CI now :)

radtriste:
When running the command `python scripts/manage-kogito-version.py --bump-to 1.0.0 --apps-branch 0.23.x`, it is creating a `common.cpython-38.pyc` file. Can we have a cleanup somehow ?

spolti:
it will be good to have, @Kaitou786 can you please address it?

Kaitou786:
Yeah, sure let me check this

Kaitou786:
We need to define a enviornment variable `export PYTHONDONTWRITEBYTECODE=1` so it doesn't write the byte code. I tried defining as a part of the script with `os.environ` but it doesn't work. If we directly define on shell and then run the python code. It doesn't create the files. Maybe we can write this part in `README.md` @spolti @radtriste wdyt?

Kaitou786:
Now the files are not creating :)

radtriste (approved):

spolti:
cool, thanks.
let's wait the CI report.

PR #158, opened 2020-05-13
[KOGITO-2116][KOGITO 2149] Adjust deploy pipeline and Setup promote pipeline

radtriste:
https://issues.redhat.com/browse/KOGITO-2149
https://issues.redhat.com/browse/KOGITO-2116

`Jenkinsfile-Nightly` will be deprecated and replaced by `Jenkinsfile.deploy` (which is a copy of it with some improvements)
With this PR, we will need to update the Jenkins job with the correct filename.
Finally default deployment registry is done Openshift.

The parent nightly pipeline will first call the `Jenkinsfile.deploy` pipeline and then if everything is ok so after operator tests, will call the `Jenkinsfile.promote` pipeline.
See https://github.com/kiegroup/kogito-pipelines/pull/5

### DIFF
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -1,0 +1,245 @@
+//Checks master branch for updates every day
+//If found triggers the pipeline, 
+//The pipeline is set to run between 12:00AM - 3:00AM UTC
+
+IMAGES = ["kogito-quarkus-ubi8", 
+            "kogito-quarkus-jvm-ubi8",
+            "kogito-quarkus-ubi8-s2i",
+            "kogito-springboot-ubi8",
+            "kogito-springboot-ubi8-s2i",
+            "kogito-data-index",
+            "kogito-jobs-service",
+            "kogito-management-console"]
+
+pipeline {
+    agent { label 'jenkins-slave' }
+    parameters {
+        string(name: 'DISPLAY_NAME', defaultValue: '', description: 'Setup a specific build display name')
+        string(name: 'BUILD_BRANCH_NAME', defaultValue: 'master', description: 'Which branch to build ? Set if you are not on a multibranch pipeline.')
+
+        string(name: 'IMAGE_REGISTRY_CREDENTIALS', defaultValue: '', description: 'Image registry credentials to use to deploy images. Will be ignored if no IMAGE_REGISTRY is given')
+        string(name: 'IMAGE_REGISTRY', defaultValue: '', description: 'Image registry to use to deploy images')
+        string(name: 'IMAGE_NAMESPACE', defaultValue: 'kiegroup', description: 'Image namespace to use to deploy images')
+        string(name: 'IMAGE_NAME_SUFFIX', defaultValue: '', description: 'Image name suffix to use to deploy images. In case you need to change the final image name, you can add a suffix to it.')
+        string(name: 'IMAGE_TAG', defaultValue: '', description: 'Image tag to use to deploy images')
+    }
+
+    // Keep commented if no env var is defined
+    // environment {
+        // Keep commented. It is for documentation only
+        // Cannot define here as an env variable defined here is immutable ...
+        // BRANCH_NAME => should be set by the multibranch pipeline. If single pipeline, then it is set with value from ${params.BUILD_BRANCH_NAME}
+
+        // OPENSHIFT_API => Taken from Jenkins global env
+        // OPENSHIFT_REGISTRY_ROUTE => Taken from Jenkins global env
+
+        // DEPLOY_IMAGE_REGISTRY_CREDENTIALS
+        // DEPLOY_IMAGE_USE_OPENSHIFT
+        // DEPLOY_IMAGE_REGISTRY
+        // DEPLOY_IMAGE_NAMESPACE
+        // DEPLOY_IMAGE_NAME_SUFFIX
+        // DEPLOY_IMAGE_TAG
+    // }
+
+    stages {
+        stage('Initialization') {
+            steps {
+                script {
+                    sh "docker rm -f \$(docker ps -a -q) || docker rmi -f \$(docker images -q) || date"
+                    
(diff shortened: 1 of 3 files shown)


sutaakar (approved):

Kaitou786:
Do we plan to use the `Jenkinsfile.deploy` for PR checks?

Jenkinsfile.deploy
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+        stage('Prepare offline kogito-examples') {
+            steps {
+                sh "make clone-repos"
+            }
+        }
+        stage('Build and test kogito-quarkus-ubi8 image') {
+            steps {
+                sh "make kogito-quarkus-ubi8"
```
Sgitario, line 101: How long these stages used to take? (I mean the kogito-quarkus-ubi8, kogito-quarkus-jvm-ubi8, kogito-springboot-ubi8, etc etc). 
If long, maybe it's a good idea to run these stages in parallel:

```
stage('Build components) {
   parallel {
      stage('Build and test kogito-quarkus-ubi8 image') {
           ...
      }
     
      stage('Build and test kogito-quarkus-jvm-ubi8 image') {
            ...
      }

      ...
   }
}
```

Jenkinsfile.deploy
```diff
@@ -0,0 +1,220 @@
...
+    post {
+        always {
+            junit testResults: 'target/test/results/*.xml', allowEmptyResults: true
+        }
+    }
+}
+
+void tagImages() {
+    for(String imageName : IMAGES) {
+        sh "docker tag quay.io/kiegroup/${imageName}:latest ${buildImageName(imageName)}"
```
Sgitario, line 177: Using ":latest" is dangerous as concurrent builds might occur. As the latest tag is set by the make commands, we can't easily change this, therefore we should disable concurrent builds in the pipeline by doing:

```
options { disableConcurrentBuilds() }
```

This way, I think we're safe.

Jenkinsfile.deploy
```diff
@@ -0,0 +1,220 @@
...
+        string(name: 'IMAGE_NAME_SUFFIX', defaultValue: '', description: 'Image name suffix to use to deploy images. In case you need to change the final image name, you can add a suffix to it.')
+        string(name: 'IMAGE_TAG', defaultValue: '', description: 'Image tag to use to deploy images')
+    }
+    stages {
+        stage('Initialization') {
+            steps {
+                script {
+                    sh "docker rm -f \$(docker ps -a -q) || docker rmi -f \$(docker images -q) || date"
+                    
+                    setupImageEnvVarsFromParams("","DEPLOY")
```
Sgitario, line 50: As far I can see, there are some environment variables within the pipeline:
OPENSHIFT_API
OPENSHIFT_REGISTRY_ROUTE
DEPLOY_IMAGE_NAME_SUFFIX
DEPLOY_IMAGE_REGISTRY
DEPLOY_IMAGE_NAMESPACE
DEPLOY_IMAGE_TAG

I think it's good practice to declare them within the "environment" block at pipeline level (or stage level if only applies to one stage).
Also, these variables can be set dynamically, so we might not need to run this method.
See more in https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables.

Sgitario (commented):
Some minor comments.

Jenkinsfile.deploy
radtriste: This were the original stages, I did not change that.
Not sure the cekit does support parallel execution on same folder. 
I would leave for later improvements and not part of this issue.
Created https://issues.redhat.com/browse/KOGITO-2226

Jenkinsfile.deploy
radtriste: Those names are generated by the cekit builder.
The builds are already concurrent and cannot run on the same node so we are safe already.
I wouldn't use `disableConcurrentBuilds()` because it may block whereas it could be executed in parallel on different nodes

Jenkinsfile.deploy
radtriste: good point I will declare them commented so we know they are used, but I won't declare them uncommented as in this case, they become immutable ...

And I did not put the dynamic set because if you check the `setupImageEnvVarsFromParams` you will see that it is a bit more complex and it would become unreadable ...
And I wanted to set them all in one place so we are more consistent.

radtriste:
@Kaitou786 no plan yet but we could investigate on that later. Can you please create an issue into the epic https://issues.redhat.com/browse/KOGITO-1871 ?

Kaitou786:
Shouldn't we remove `Jenkinsfile-Nightly` as we are going to use deploy and promote for nightlies?

radtriste:
`Jenkinsfile-Nightly will be deprecated` (in the description)

I keep it for now for a smooth transition and a fallback if needed ;)

Kaitou786:
Okay ;)

Kaitou786 (approved):

Sgitario (approved):

Jenkinsfile.deploy
spolti: +1

Jenkinsfile.deploy
spolti: I think it worth a try.

spolti (approved):

Jenkinsfile.deploy
Kaitou786: I think Cekit doesn't support parallel execution on the same folder
See: https://jenkins-kogito-tools.apps.kogito.automation.rhmw.io/blue/organizations/jenkins/Labs%2Fimages-parallel-test/detail/images-parallel-test/5/pipeline/     
Only one of the tests proceeds while others fail. See different builds for this same pipeline

spolti: Can we choose which pipeline item will be run in parallel?

Kaitou786: Yes, we can decide which stages to run in parallel

PR #167, opened 2020-05-27
[KOGITO-2226] - Kogito Images pipeline builds & tests in parallel

Kaitou786:
See: https://issues.redhat.com/browse/KOGITO-2226

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,13 +6,25 @@
 def buildNumber = env.BUILD_NUMBER as int
 if (buildNumber > 1) milestone(buildNumber - 1)
 milestone(buildNumber)
+IMAGES = ["kogito-quarkus-ubi8", 
+            "kogito-quarkus-jvm-ubi8",
+            "kogito-quarkus-ubi8-s2i",
+            "kogito-springboot-ubi8",
+            "kogito-springboot-ubi8-s2i",
+            "kogito-data-index",
+            "kogito-jobs-service",
+            "kogito-management-console"]
 
 pipeline{
     agent { label 'jenkins-slave'}
     stages{
         stage('Initialization'){
             steps{
-                sh "docker rm -f \$(docker ps -a -q) || docker rmi -f \$(docker images -q) || date"
+                script{
+                    cleanWorkspaces()
+                }
+                sh "docker rm -f \$(docker ps -a -q) || date"
+                sh "docker rmi -f \$(docker images -q) || date"
             }
         }
         stage('Validate CeKit Image and Modules descriptors'){
@@ -41,44 +53,26 @@ pipeline{
                 sh "make clone-repos"
             }
         }
-        stage('Build and test kogito-quarkus-ubi8 image'){
+        stage('Build and Test Images'){
             steps{
-                sh "make kogito-quarkus-ubi8"
-            }
-        }
-        stage('Build and test kogito-quarkus-jvm-ubi8 image'){
-            steps{
-                sh "make kogito-quarkus-jvm-ubi8"
-            }
-        }
-        stage('Build and test kogito-quarkus-ubi8-s2i image'){
-            steps{
-                sh "make kogito-quarkus-ubi8-s2i"
-            }
-        }
-        stage('Build and test kogito-springboot-ubi8 image'){
-            steps{
-                sh "make kogito-springboot-ubi8"
-            }
-        }
-        stage('Build and test kogito-springboot-ubi8-s2i image '){
-            steps{
-                sh "make kogito-springboot-ubi8-s2i"
-            }
-        }
-        stage('Build and test kogito-data-index image '){
-            steps{
-                sh "make kogito-data-index"
-            }
-        }
-        stage('Build and test kogito-jobs-service image '){
-            steps{
-                sh "make kogito-jobs-service"
-            }
-        }
-        stage('Build and test kogito-management-console image '){
-            steps{
-                sh "make kogito-management-console"
+                script {
+                    build_stages = [:]
+                    IMAGES.each{ image -> build_stages["${image}"] = {
+                            createWorkspace("$image")
+                            copyWorkspace("$image")
+                            dir(getWorkspacePath("$image")){
+                                try{
+                                    sh "make ${image}"
+                                }
+                                finally{
+                                    junit 'target/test/results/*.xml'
+                                }
+                            
+                            }
+                        }
+                    }
+                    parallel build_stages
+                }
             }
         }
         stage('Finishing'){
@@ -89,8 +83,27 @@ pipeline{
     }
     post{
         always{
-            junit 'target/test/results/*.xml'
+            script{
+                cleanWorkspaces()
+            }
         }
     }
 }
 
+
+
+void  createWorkspace(String image){
+    sh "mkdir -p ${getWorkspacePath(image)}"
+}
+void copyWorkspace(String image){
+    sh "rsync -av --progress . ${getWorkspacePath(image)} --exclude workspaces"
+}
+void cleanWorkspaces(){
+    sh "rm -rf ${getWorkspacesPath()}"
+}
+String getWorkspacesPath(){
+    return "${WORKSPACE}/workspaces"
+}
+String getWorkspacePath(String image){
+    return "${getWorkspacesPath()}/${image}"
+}
--- a/Jenkinsfile-Nightly
+++ b/Jenkinsfile-Nightly
@@ -2,6 +2,15 @@
 //If found triggers the pipeline, 
 //The pipeline is set to run between 12:00AM - 3:00AM UTC
 
+IMAGES = ["kogito-quarkus-ubi8", 
+            "kogito-quarkus-jvm-ubi8",
+            "kogito-quarkus-ubi8-s2i",
+            "kogito-springboot-ubi8",
+            "kogito-springboot-ubi8-s2i",
+            "kogito-data-index",
+            "kogito-jobs-service",
+            "kogito-management-console"]
+
 pipeline{
     agent { label 'jenkins-slave'}
      triggers{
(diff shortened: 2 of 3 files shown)


Kaitou786:
rebuild

Jenkinsfile
```diff
@@ -36,49 +47,61 @@ pipeline{
...
+        stage('Build and Test Images'){
+            parallel{
+                stage('Build and test kogito-quarkus-ubi8 image'){
+                    steps{
+                        sh "make kogito-quarkus-ubi8"
+                    }
+                }
+                stage('Build and test kogito-quarkus-jvm-ubi8 image'){
+                    steps{
+                        sh "cd /app/jenkins/workspace/kogito-quarkus-jvm-ubi8 && make kogito-quarkus-jvm-ubi8"
```
radtriste: use [dir](https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#dir-change-current-directory) instead of `cd` in command

radtriste: Also please use `${WORKSPACE}` instead of `/app/jenkins/workspace`, that way we are more generic and that will be easier to move the job to another Jenkins

Jenkinsfile
```diff
@@ -7,11 +7,22 @@ def buildNumber = env.BUILD_NUMBER as int
...
+            "kogito-jobs-service",
+            "kogito-management-console"]
+
 pipeline{
     agent { label 'jenkins-slave'}
     stages{
         stage('Initialization'){
             steps{
+                script{
+                    cleanWorkspaces()
```
radtriste, line 24: Add it into the `always` post step

Jenkinsfile
```diff
@@ -93,4 +116,18 @@ pipeline{
...
+    }
+}
+void createWorkspaces() {
+        for(String imageName : IMAGES){
+        sh "mkdir -p /app/jenkins/workspace/$imageName"
+    }
+}
+void copyWorkspaces(){
+   for(String imageName : IMAGES){
+        sh "cp -R . /app/jenkins/workspace/$imageName"
```
radtriste: Could we put this into a specific `workspaces` folder, to avoid any conflict ?

Jenkinsfile
```diff
@@ -36,49 +47,61 @@ pipeline{
...
             steps{
-                sh "make kogito-jobs-service"
+                sh "make clone-repos"
             }
         }
-        stage('Build and test kogito-management-console image '){
-            steps{
-                sh "make kogito-management-console"
+        stage('Build and Test Images'){
+            parallel{
```
radtriste: Maybe we could do that a bit more generic inside this stage.
The parallel stages could be generated dynamically for each image in `IMAGES`.
And I would move the create/copy workspace directly in there.
This is already explicitly defined what we are doing here, so it is fine to do a bit of scripted pipeline IMO.
wdyt ?

radtriste (commented):
I know this is still draft but can you have a look to these comments ?

Jenkinsfile
Kaitou786: Okay, makes sense. Right now just testing

Jenkinsfile
Kaitou786: Okay, I will add it to `always` post section though I think it would be good if we put it in the start as well

Jenkinsfile
Kaitou786: Yeah makes sense.

Jenkinsfile
Kaitou786: I'm currently testing for creating dynamic stages, but looks like we'll have to use scripted pipeline, we can check if it does not decrease any readability we can consider using it.

radtriste: scripted pipeline should be used only in this stage, aka `'Build and Test Images'` stage.
The rest stays like it is.

spolti (approved):
besides Tristan's comments, sounds good to me.
nice work :)

Jenkinsfile
```diff
@@ -7,11 +7,22 @@ def buildNumber = env.BUILD_NUMBER as int
 if (buildNumber > 1) milestone(buildNumber - 1)
 milestone(buildNumber)
 
+IMAGES = [  "kogito-quarkus-jvm-ubi8",
```
sutaakar: Seems to miss `kogito-quarkus-ubi8`

spolti: good catch :)

Kaitou786: Thanks :)

sutaakar (approved):

Jenkinsfile
```diff
@@ -6,12 +6,25 @@
...
+IMAGES = ["kogito-quarkus-ubi8", 
+            "kogito-quarkus-jvm-ubi8",
+            "kogito-quarkus-ubi8-s2i",
+            "kogito-springboot-ubi8",
+            "kogito-springboot-ubi8-s2i",
+            "kogito-data-index",
+            "kogito-jobs-service",
+            "kogito-management-console"]
+
+IMAGES_WS = "/app/jenkins/workspace/workspaces/"   //Jenkins_ws/workspaces  Temporary directory will be created under this directory
```
radtriste: Isn't `/app/jenkins/workspace` equivalent to the existing Jenkins env var `WORKSPACE` ?

I'd rather not having hardcoded path here ... we could use the `WORKSPACE` env var

Jenkinsfile
```diff
@@ -89,8 +85,21 @@ pipeline{
...
+                cleanWorkspaces()
+            }
         }
     }
 }
 
+
+
+void  createWorkspace(String image){
+    sh "mkdir -p $IMAGES_WS$image/"
```
radtriste: Please use `${VAR}` for readibility

Jenkinsfile
```diff
@@ -89,8 +85,21 @@ pipeline{
...
+}
+void copyWorkspace(String image){
+    sh "cp -Rap * $IMAGES_WS$image/"
```
radtriste: Instead of repeating the path here also, can you extract it in a `getImageWorkspacePath` method ?

Jenkinsfile
radtriste: also that could be provided by a `getWorkpspacesPath` method instead of a global variables IMO

radtriste (changes requested):

Jenkinsfile
Kaitou786: The env variable workspace contains the path where the current job is executing, which is different than `/app/jenkins/workspace`

radtriste: is there a reason why we moved that to another place ?

Kaitou786: Didn't understand wdym?

Jenkinsfile
```diff
@@ -41,44 +52,28 @@ pipeline{
...
-        stage('Build and test kogito-management-console image '){
-            steps{
-                sh "make kogito-management-console"
+                script {
+                    tests = [:]
+                    IMAGES.each{ image -> tests["${image}"] = {
+                            createWorkspace("${image}")
+                            copyWorkspace("${image}")
+                            path = getWorkspacePath("${image}")
+                            dir("${path}"){
```
radtriste:
```suggestion
                            dir(getWorkspacePath(image)){
```

Jenkinsfile
```diff
@@ -41,44 +52,28 @@ pipeline{
...
+                                try{
+                                    sh "make ${image}"
+                                }
+                                finally{
+                                    step([$class: 'JUnitResultArchiver', testResults: 'target/test/results/*.xml', healthScaleFactor: 1.0])
+                                    junit 'target/test/results/*.xml'
```
radtriste: why are we doing twice the same thing ?

Jenkinsfile
```diff
@@ -89,8 +84,26 @@ pipeline{
...
+            }
         }
     }
 }
 
+
+
+void  createWorkspace(String image){
+    String path = getWorkspacePath("${image}")
+    sh "mkdir -p ${path}"
```
radtriste:
```suggestion
    sh "mkdir -p ${getWorkspacePath(image)}"
```

Jenkinsfile
```diff
@@ -89,8 +84,26 @@ pipeline{
...
+}
+void copyWorkspace(String image){
+    String path = getWorkspacePath("${image}")
+    sh "rsync -av --progress . ${path} --exclude workspaces"
```
radtriste:
```suggestion
    sh "rsync -av --progress . ${${getWorkspacePath(image)}} --exclude workspaces"
```

Jenkinsfile
```diff
@@ -89,8 +84,26 @@ pipeline{
...
+}
+void cleanWorkspaces(){
+    sh "rm -rf ${WORKSPACE}/workspaces/"
+}
+String getWorkspacePath(String image){
+    return "${WORKSPACE}/workspaces/${image}"
+}
```
radtriste:
```suggestion
void cleanWorkspaces(){
    sh "rm -rf ${getWorkspacesPath()}"
}
String getWorkspacesPath(){
    return "${WORKSPACE}/workspaces/"
}
String getWorkspacePath(String image){
    return "${getWorkspacesPath()}/${image}"
}
```

radtriste (changes requested):

Jenkinsfile
```diff
@@ -41,44 +52,26 @@ pipeline{
...
-        stage('Build and test kogito-jobs-service image '){
-            steps{
-                sh "make kogito-jobs-service"
-            }
-        }
-        stage('Build and test kogito-management-console image '){
-            steps{
-                sh "make kogito-management-console"
+                script {
+                    tests = [:]
```
radtriste: maybe rename to `build_stages`

Jenkinsfile
```diff
@@ -89,8 +82,27 @@ pipeline{
...
+
+
+void  createWorkspace(String image){
+    sh "mkdir -p ${getWorkspacePath(image)}"
+}
+void copyWorkspace(String image){
+    sh "rsync -av --progress . ${getWorkspacePath(image)} --exclude workspaces"
+}
+void cleanWorkspaces(){
+    sh "rm -rf ${WORKSPACE}/workspaces/"
```
radtriste: please use `getWorkspacesPath` method

radtriste (changes requested):
2 changes that should be applied to all 3 files

Jenkinsfile
Kaitou786: Renamed

Jenkinsfile
Kaitou786: changed

radtriste (approved):
lgtm

sutaakar (approved):

spolti (approved):

PR #173, opened 2020-06-04
[KOGITO-2374] Setup Maven repo url

radtriste:
https://issues.redhat.com/browse/KOGITO-2374

Consider the given repo-url in local and remote builds.

- Refactored a bit the way scripts are updating version / env
- `KOGITO_MAVEN_REPO_URL` is added if needed to change the default Kogito artifacts url (`https://repository.jboss.org/nexus/content/groups/public/`).
- `clone-repos.sh` is using the `configure` & `settings.xml` from kogito-maven


sutaakar:
I am not sure that we should use `MAVEN_REPO_URL` for our JBoss repo.
What will happen in case user will have some domain artifacts built in his own repository and will want to use both JBoss (for common Kogito artifacts) and his own repository in the build?

radtriste:
Good point but the user should use normally the Maven central artifacts by default and not the JBoss ones or ?
Jboss one should be used only for snapshot then.

Another concern is that if we have a staging repository or another repository containing the latest snapshot artifacts and JBoss repo is not yet updated (a mirror repository takes some time for the update...), who has precedence on the other ?
Will I get the latest snapshots ?
Or we should have a way to add and override a KOGITO_MAVEN_URL env ?
and leave the MAVEN_REPO_URL free ?

Another way would be to leverage the `MAVEN_REPOS` extension

spolti:
I am not sure if remove the default repo is a good idea, maybe use -U to force download the latests artifacts?

radtriste:
@spolti yes and no because if I have 2 repositories both with those snapshots, the first one tried and answering ok will give the snapshots, and may be not up to date (mirroring in Maven can take time...)
But maybe I can define a KOGITO_MAVEN_URL with JBoss default which could be overriden ?

btw, for productive images, shouldn't the default repository be Maven central ?

spolti:
maybe disable the snapshots on jboss central might be better.

> btw, for productive images, shouldn't the default repository be Maven central ?

not really, jboss repo has been in use since beginning cloud effort, if there's a strong reason to change that, we can.

> But maybe I can define a KOGITO_MAVEN_URL with JBoss default which could be overriden ?

Even better, we can remove the jboss central from settings.xml and, during build time, if no repo is set at all, we define the default one. this way, no other repo will be added if you correctly set the maven configuration envs

radtriste:
so will go for being able to change default Maven repo via `KOGITO_MAVEN_REPO_URL` and `MAVEN_REPO_URL` keeps its old behavior.

radtriste:
@spolti @sutaakar update to `KOGITO_MAVEN_REPO_URL` done. let me know what you think

modules/kogito-maven/3.6.x/module.yaml
```diff
@@ -29,8 +29,11 @@ envs:
   - name: "MAVEN_DOWNLOAD_OUTPUT"
     description: "If set to true will print the transfer logs for downloading/uploading of maven dependencies. Defaults to false"
     example: "true"
+  - name: "KOGITO_MAVEN_REPO_URL"
```
sutaakar: Wouldn't it have more sense to name it something like JBOSS_MAVEN_REPO_URL ?
As it contains also other artifacts than just Kogito ones.

sutaakar (approved):
Looks good, one suggestion above.

modules/kogito-maven/3.6.x/module.yaml
radtriste: will rename it

radtriste:
@spolti @Kaitou786 can you please review ?

Kaitou786:
@radtriste shouldn't we also change the  `update-artifacts.py` → `update-maven-information.py` in `Jenkinsfile-Nightly` I know it is deprecated but we are also keeping it in place as the whole nightly is a bit unstable

radtriste:
you are right. Missed it. I will update it.

radtriste:
@Kaitou786 updated, as well as the README

Kaitou786:
Thanks :)

Kaitou786 (approved):

spolti (approved):

PR #181, opened 2020-06-15
[KOGITO-2445] maven repo added to tests only

radtriste:
https://issues.redhat.com/browse/KOGITO-2445

The feature tests are updated with the given artifactory as it is the one we want to test against.
The env var is no more set directly into the module directly else it will be packaged with the image as default repository when building ...

and added option to skip tests in deploy

- [ ] Pull Request does not include fixes for issues other than the main ticket

### DIFF
--- a/scripts/update-maven-information.py
+++ b/scripts/update-maven-information.py
@@ -15,6 +15,7 @@ import common
 import xml.etree.ElementTree as ET
 import requests
 import subprocess as sp
+import re
 import os
 import argparse
 from ruamel.yaml import YAML
@@ -24,6 +25,9 @@ MAVEN_MODULE="kogito-maven/3.6.x"
 DEFAULT_REPO_URL = "https://repository.jboss.org/nexus/content/groups/public/"
 DEFAULT_VERSION = "8.0.0-SNAPSHOT"
 DEFAULT_ARTIFACT_PATH = "org/kie/kogito"
+
+# behave tests that needs to be update
+S2I_BEHAVE_TESTS = {"kogito-quarkus-ubi8-s2i.feature", "kogito-springboot-ubi8-s2i.feature"}
 
 Modules = {
     #service-name: module-name(directory in which module's module.yaml file is present)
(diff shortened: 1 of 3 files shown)


tests/features/kogito-springboot-ubi8-s2i.feature
```diff
@@ -31,6 +31,7 @@ Feature: kogito-springboot-ubi8-s2i image tests
 
   Scenario: Verify if the s2i build is finished as expected and if it is listening on the expected port, test uses custom properties file to test the port configuration.
     Given s2i build /tmp/kogito-examples from process-springboot-example using master and runtime-image quay.io/kiegroup/kogito-springboot-ubi8:latest
+      | variable | value |
```
radtriste, line 35: added those to make the replacement needed much easier.
This has no impact on the tests

sutaakar: Can you please add a comment there mentioning that this line is just a placeholder?
Otherwise it won't be clear for casual reader.

spolti: 'variable | value' are the reader, the envs should be added below it.

scripts/update-maven-information.py
```diff
@@ -104,12 +108,36 @@ def update_artifacts(service,modulePath):
...
+    Updates S2I feature tests with the given repository URL.
+    :param repo_url: Maven Repository URL to set
+    '''
+    base_dir = 'tests/features'
+    pattern_branch = re.compile('\|\s*variable[\s]*\|[\s]*value[\s]*\|')
+
+    for feature in S2I_BEHAVE_TESTS:
+        print("Updating feature {0}".format(feature))
+        with open(os.path.join(base_dir, feature)) as fe:
+            updated_value = pattern_branch.sub("| variable | value |\n      | JBOSS_MAVEN_REPO_URL | {} |\n      | MAVEN_DOWNLOAD_OUTPUT | true |".format(repo_url), fe.read())
```
spolti: would be better just remove the variable | name from feature file and add the actual env name and replace it there.

spolti (changes requested):

tests/features/kogito-springboot-ubi8-s2i.feature
radtriste: There is no envs defined by default for this test. that is why we just have the `reader`

scripts/update-maven-information.py
radtriste: what do you mean ?

For now I am looking for the `| variable | value |` readers and then add the `JBOSS_MAVEN_REPO_URL` the line after (but python offers only replace so I need to add back the `| variable | value |` ...)

tests/features/kogito-springboot-ubi8-s2i.feature
radtriste: added the comment

scripts/update-maven-information.py
radtriste: by default, we don't want the `JBOSS_MAVEN_REPO_URL` env var to be defined

sutaakar (approved):

spolti (approved):

PR #209, opened 2020-08-10
KOGITO-3013 - add monitoring dashboards scripts and tests

r00ta:
Jira: https://issues.redhat.com/browse/KOGITO-3049

This PR aims to extract the monitoring dashboards that are generated by a kogito application that uses the `monitorin-addon`. These dashboards are then copied inside the container and a script that is run inside the container will update the config map with the dashboards. Basically I'm re-using the same appoach that was implemented for the persistence/protobuf files. 

To recap, the architecture we'd like to build looks like the following
RuntimeMetrics

In particular the prometheus+grafana operators will be installed, and the dashboards generated at build time have to be injected into grafana. 

- [x] You've tested the new feature/bug fix in an actual OpenShift cluster


r00ta:
Jenkins test this

modules/kogito-monitoring/added/kogito-monitoring.sh
```diff
@@ -0,0 +1,105 @@
...
+# with the generated dashboards files mounted in the file system.
+# Can be called multiple times or outside of a k8s cluster.
+# If outside the cluster, just skips the update
+function update_monitoring_configmap() {
+    if ! is_running_on_kubernetes; then
+        log_info "---> [monitoring] Not running on kubernetes cluster, skipping config map update"
+        return 0
+    fi
+
+    local config_map=$(cat $KOGITO_HOME/podinfo/monitoringcm)
```
r00ta, line 60: Dumb question: who should create the config maps in `podinfo`? I see that for the persistence the file is called `protobufcm`, is it something provided by kubernetes or what?

ricardozanini: You need to have the operator to  create the configMap for you. This approach won't work if it's not there. See: https://docs.google.com/presentation/d/1Ji-Ob4KGU4j7rcxUMD_rp3l5xBfJAqmood4yZchEnFk/edit#slide=id.g8f45d2793a_0_45 Slide 6

And also have the Deployment to mount the podinfo for you as well.

ricardozanini (changes requested):
Hi! 
Many thanks for this PR, would be really nice to have dashboards provided by the Operator. I have a couple questions, tho:

1. Have you tested this script on an actual Kube cluster? I ask because I'm curious to see how this configMap gets updated with the dashboard and how the files are structured in there. 
2. These functions are almost the same we implemented in the protobuf use case. Do you mind refactoring to a common module and reuse them across both use cases: protobuf and dashboards? Will be easier to maintain.
3. You should also send a PR to the Operator to create the CM for you as we create the protobuf CM. Should be the CMs and not only one (like `myservice-files`) since the protobuf one needs to be mounted in the DataIndex as well, it's a different approach. What are you going to do with the information inside this CM? You will mount it in a Grafana resource provided by their operator? Would you give the CM for them to create the dashboard? What's the outcome of having this CM in the namespace?

Thanks, sorry for begin a PITA. :(

modules/kogito-monitoring/added/kogito-monitoring.sh
```diff
@@ -0,0 +1,105 @@
...
+
+    if [ "${data}" != "" ]; then
+        data="${data%,}"
+    fi
+
+    body="[ { \"op\": \"replace\", \"path\": \"/metadata/annotations\", \"value\": { ${metadata} } }, { \"op\": \"replace\", \"path\": \"/data\", \"value\": { ${data} } } ]"
+    log_info "---> [monitoring] About to patch configMap ${config_map}"
+    # prints the raw data
+    echo "Body: ${body}"
+    printf "%s" "${body}" >$KOGITO_HOME/data/dashboards/configmap_patched.json
```
spolti, line 94: we could mock these requests using a little python server + bats tests.
here is an example:
https://github.com/jboss-container-images/jboss-kie-modules/blob/master/jboss-kie-kieserver/tests/bats/jboss-kie-kieserver-hooks.bats
You just need to have a real request to copy body and response to mock it.


@ricardozanini @radtriste 
we could move this mock server to a common repo and reuse, wdyt?

spolti: PLus, instead duplicating code we could have a shared module and reuse it for protobuff and this new stuff.

ricardozanini: ^ READ MY COMMENTS! :laughing:

spolti:
@r00ta Hi, what is the status of this issue?

spolti:
@r00ta what is the status of this?

spolti:
the jira is closed, can we close it as well?

PR #253, opened 2020-10-06
[KOGITO-3553] - Ignore Self signed certificates from maven while bui…

Kaitou786:
…lding images

The PR aims to provide a mechanism to relax the SSL for user-generated self-signed certificates, Currently, our image build fails when a user has a maven mirror with self-signed certificates, after the PR setting `MAVEN_IGNORE_SELF_SIGNED_CERTIFICATE` to "true" relax the constraint making our image to build successfully with self-signed maven mirror's certificate. For more info please [see](https://stackoverflow.com/questions/21252800/how-to-tell-maven-to-disregard-ssl-errors-and-trusting-all-certs)
See: https://issues.redhat.com/browse/KOGITO-3553

Signed-off-by: Tarun Khandelwal <tarkhand@redhat.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,6 +37,11 @@ pipeline{
                     }
 
                     githubscm.checkoutIfExists('kogito-images', changeAuthor, changeBranch, 'kiegroup', changeTarget, true)
+
+                    //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
+                    if(env.MAVEN_MIRROR_URL != ""){
+                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert=True"
+                    }
                 }
             }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,8 +39,8 @@ pipeline{
                     githubscm.checkoutIfExists('kogito-images', changeAuthor, changeBranch, 'kiegroup', changeTarget, true)
 
                     //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
-                    if(env.MAVEN_MIRROR_URL != ""){
-                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert=True"
+                    if(env.MAVEN_MIRROR_URL != ''){
+                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert"
                     }
                 }
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline{
 
                     //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
                     if(env.MAVEN_MIRROR_URL != ''){
-                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert"
+                        sh 'python3 scripts/update-tests.py --ignore-self-signed-cert'
                     }
                 }
             }
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -215,7 +215,7 @@ pipeline {
 
                     //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
                     if(env.MAVEN_MIRROR_URL != ''){
-                        updateTestsCommand += " --ignore-self-signed-cert"
+                        updateTestsCommand += ' --ignore-self-signed-cert'
                     }
                     // Launch update tests
                     sh updateTestsCommand
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -206,7 +206,7 @@ pipeline {
                     }
 
                     // Set kogito-examples to bot author/branch if release
-                    if(params.EXAMPLES_REF != ""){
+                    if(params.EXAMPLES_REF != ''){
                         updateTestsCommand += " --examples-ref ${params.EXAMPLES_REF}"
                     }
                     if(params.EXAMPLES_URI){
@@ -214,8 +214,8 @@ pipeline {
                     }
 
                     //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
-                    if(env.MAVEN_MIRROR_URL != ""){
-                        updateTestsCommand += --ignore-self-signed-cert=True
+                    if(env.MAVEN_MIRROR_URL != ''){
+                        updateTestsCommand += " --ignore-self-signed-cert"
                     }
                     // Launch update tests
                     sh updateTestsCommand
--- a/Jenkinsfile.deploy
+++ b/Jenkinsfile.deploy
@@ -213,6 +213,10 @@ pipeline {
                         updateTestsCommand += " --examples-uri ${params.EXAMPLES_URI}"
                     }
 
+                    //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
+                    if(env.MAVEN_MIRROR_URL != ""){
+                        updateTestsCommand += --ignore-self-signed-cert=True
+                    }
                     // Launch update tests
                     sh updateTestsCommand
 
--- a/modules/kogito-maven/3.6.x/added/configure-maven.sh
+++ b/modules/kogito-maven/3.6.x/added/configure-maven.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+#Please keep them in alphabatical order
 function prepareEnv() {
     unset HTTP_PROXY_HOST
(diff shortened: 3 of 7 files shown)


vaibhavjainwiz (approved):
LGTM

modules/kogito-maven/3.6.x/added/configure-maven.sh
```diff
@@ -19,12 +19,14 @@ function prepareEnv() {
     unset MAVEN_REPO_SNAPSHOTS_CHECKSUM_POLICY
     unset MAVEN_REPO_URL
     unset MAVEN_REPOS
+    unset IGNORE_SELF_SIGNED_CERTIFICATE
```
radtriste: thinking that we should maybe prefix it with `MAVEN` as this is maven particularity.
wdyt ?

modules/kogito-maven/3.6.x/module.yaml
```diff
@@ -62,7 +62,9 @@ envs:
   - name: "MAVEN_REPOS"
     description: "Used to define multiple repositories, this env defines a prefix that will be used to create different repositories."
     example: "CENTRAL,INTERNAL"
-
+  - name: "IGNORE_SELF_SIGNED_CERTIFICATE"
+    description: "When set, use of relaxed SSL check for user generated certificates. Default value is false"
+    value: "true"
```
radtriste: `Default value is false` and then `value: "true"` :)
just put an example and not a value

Kaitou786: Ah yes, :p, missed this, will change it

modules/kogito-maven/3.6.x/added/configure-maven.sh
Kaitou786: It is part of, maven module but I think as other vars have that as a prefix I'll add it to this one :)

modules/kogito-maven/3.6.x/added/configure-maven.sh
```diff
@@ -19,12 +19,14 @@ function prepareEnv() {
     unset MAVEN_REPO_SNAPSHOTS_CHECKSUM_POLICY
     unset MAVEN_REPO_URL
     unset MAVEN_REPOS
+    unset MAVEN_IGNORE_SELF_SIGNED_CERTIFICATE
```
spolti: please keep in alphabetical order.

modules/kogito-maven/3.6.x/added/configure-maven.sh
```diff
@@ -19,12 +19,14 @@ function prepareEnv() {
...
 }
 
 function configure() {
     configure_proxy
     configure_mirrors
     configure_maven_download_output
+    configure_self_signed_certificates
```
radtriste:
```suggestion
    configure_maven_self_signed_certificates
```

modules/kogito-maven/3.6.x/added/configure-maven.sh
```diff
@@ -92,6 +94,12 @@ function configure_maven_download_output() {
     fi
 }
 
+function configure_self_signed_certificates() {
```
radtriste:
```suggestion
function configure_maven_self_signed_certificates() {
```

modules/kogito-maven/3.6.x/added/configure-maven.sh
spolti: I would change this function to ignore_self

spolti (changes requested):

spolti (approved):

radtriste:
For the pipeline, I think we would need a new possibility in `scripts/update-tests.py` to disable certificate verification and that would add `| MAVEN_IGNORE_SELF_SIGNED_CERTIFICATE | true |` into the env of feature tests (see how it is done with MAVEN_REPO_URL for example).
And then in pipeline, before launching the tests, if `env.MAVEN_MIRROR_URL` is defined, then we should update the tests with the certificate disabled.

Kaitou786:
> For the pipeline, I think we would need a new possibility in `scripts/update-tests.py` to disable certificate verification and that would add `| MAVEN_IGNORE_SELF_SIGNED_CERTIFICATE | true |` into the env of feature tests (see how it is done with MAVEN_REPO_URL for example).

Okay, will add the function, this should be part of this PR or another small JIRA?

radtriste:
I guess you can implement as part of this PR as it is related

Jenkinsfile
```diff
@@ -37,6 +37,11 @@ pipeline{
                     }
 
                     githubscm.checkoutIfExists('kogito-images', changeAuthor, changeBranch, 'kiegroup', changeTarget, true)
+
+                    //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
+                    if(env.MAVEN_MIRROR_URL != ""){
```
radtriste: please use single-quote

Jenkinsfile
```diff
@@ -37,6 +37,11 @@ pipeline{
...
+                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert=True"
```
radtriste: you should not need to have `=True`. if option is there, then it is true

scripts/common.py
```diff
@@ -222,7 +222,15 @@ def update_maven_repo_in_behave_tests(repo_url, replaceJbossRepository):
         envVarKey = "JBOSS_MAVEN_REPO_URL"
     replacement = "| variable | value |\n      | {} | {} |\n      | MAVEN_DOWNLOAD_OUTPUT | true |".format(envVarKey, repo_url)
     update_in_behave_tests(pattern, replacement)
-
+def ignore_maven_self_signed_certificate_in_behave_tests():
+    """
+    Sets the enviornment variable to ignore the self-signed certificates in maven
```
radtriste:
```suggestion
    Sets the environment variable to ignore the self-signed certificates in maven
```

scripts/update-tests.py
```diff
@@ -17,6 +17,7 @@
     parser.add_argument('--examples-uri', dest='examples_uri', help='To update the examples uri for testing')
     parser.add_argument('--examples-ref', dest='examples_ref', help='To update the examples ref for testing')
     parser.add_argument('--artifacts-version', dest='artifacts_version', help='To update the artifacts version for testing')
+    parser.add_argument('--ignore-self-signed-cert', dest='ignore_self_signed_cert', help='If set to true will relax the SSL for user-generated self-signed certificates', default=False)
```
radtriste: see `--replace-jboss-repo` option to have similar behavior

Jenkinsfile
```diff
@@ -37,6 +37,11 @@ pipeline{
                     }
 
                     githubscm.checkoutIfExists('kogito-images', changeAuthor, changeBranch, 'kiegroup', changeTarget, true)
+
+                    //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
+                    if(env.MAVEN_MIRROR_URL != ''){
+                        sh "python3 scripts/update-tests.py --ignore-self-signed-cert"
```
radtriste: as long as you are making usage of gstring (which means include a variable of a string with `${VAR}`), you should use single-quote ;)
```suggestion
                        sh 'python3 scripts/update-tests.py --ignore-self-signed-cert'
```

Jenkinsfile.deploy
```diff
@@ -206,13 +206,17 @@ pipeline {
...
+                    if(params.EXAMPLES_REF != ''){
                         updateTestsCommand += " --examples-ref ${params.EXAMPLES_REF}"
                     }
                     if(params.EXAMPLES_URI){
                         updateTestsCommand += " --examples-uri ${params.EXAMPLES_URI}"
                     }
 
+                    //Ignore self-signed certificates if MAVEN_MIRROR_URL is defined
+                    if(env.MAVEN_MIRROR_URL != ''){
+                        updateTestsCommand += " --ignore-self-signed-cert"
```
radtriste:
```suggestion
                        updateTestsCommand += ' --ignore-self-signed-cert'
```

radtriste (changes requested):
gonna test this tomorrow

Jenkinsfile
Kaitou786: Ah okay, thanks, will keep in mind from next time :)

radtriste (approved):

radtriste:
Tested ok with a specific MAVEN_MIRROR_URL with self signed certificate

PR #282, opened 2020-11-16
[KOGITO-3861] - kogito user does not have permission on home director…

spolti:
…y when running on OCP

See: https://issues.redhat.com/browse/KOGITO-3861

- [x] You've tested the new feature/bug fix in an actual OpenShift cluster

### DIFF
--- a/modules/kogito-system-user/add-user
+++ b/modules/kogito-system-user/add-user
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-groupadd -r kogito -g 1001 && useradd -u 1001 -r -g 1001 -m -d ${KOGITO_HOME} -s /sbin/nologin -c "Kogito user" ${USER}
+groupadd -r kogito -g 1001 && useradd -u 1001 -r -g root -G ${USER} -m -d ${KOGITO_HOME} -s /sbin/nologin -c "Kogito user" ${USER}
 
 mkdir ${KOGITO_HOME}/bin
-chown -R 1001:0 ${KOGITO_HOME}
+chmod ug+rwX ${KOGITO_HOME}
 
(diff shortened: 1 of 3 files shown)


Kaitou786 (approved):

modules/kogito-system-user/add-user
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-groupadd -r kogito -g 1001 && useradd -u 1001 -r -g 1001 -m -d ${KOGITO_HOME} -s /sbin/nologin -c "Kogito user" ${USER}
+groupadd -r kogito -g 1001 && useradd -u 1001 -r -g root -G ${USER} -m -d ${KOGITO_HOME} -s /sbin/nologin -c "Kogito user" ${USER}
```
ricardozanini, line 3: Is there a difference using `0` instead of `root`? Just asking, [because here](https://www.openshift.com/blog/a-guide-to-openshift-and-uids) they mention to use `0`:

> For cases where the files and directories need to be owned by the UID executing the application: 
> An approach that can be used, is to **make the files and directory writable by GID=0** and then use a script or wrapper as Entrypoint to dynamically modify the ownership of the files and directories to the allocated Namespace UID used to invoke the Pod.

ricardozanini (approved):

modules/kogito-system-user/add-user
spolti: there's no difference, root gid is 0:

```
cat /etc/group
root:x:0:
...
```

ricardozanini: yup, I know.. just wondering if this can have any caveats...

sutaakar (approved):

PR #1136, opened 2022-02-25
[1.11.x] [RHPAM-4195] Enable multi-arch for existing builder and runtime JVM images

antncl:
JIRA issue : https://issues.redhat.com/browse/RHPAM-4195

Enable ppc64le only since s390x was dropped. There are some workarounds added to graalvm configuration scripts to check the arch to avoid us having to add a whole new image because of those unsupported native packages.

Please review this and as soon as there is positive concencus I will raise the PR for main too. Mainly due to the configuration script arch detection, it's a trade off to basically avoid having a whole new image which requires Comet requests or whatever.

### DIFF
--- a/content_sets.yaml
+++ b/content_sets.yaml
@@ -18,8 +18,8 @@ x86_64:
 s390x:
   - rhel-8-for-s390x-baseos-rpms
   - rhel-8-for-s390x-appstream-rpms
-  - openj9-1-for-rhel-8-s390x-rpms
+  - rhocp-4.7-for-rhel-8-s390x-rpms
 ppc64le:
   - rhel-8-for-ppc64le-baseos-rpms
   - rhel-8-for-ppc64le-appstream-rpms
-  - openj9-1-for-rhel-8-ppc64le-rpms
+  - rhocp-4.7-for-rhel-8-ppc64le-rpms
--- a/modules/kogito-graalvm-installer/20.x-java-1.8/configure
+++ b/modules/kogito-graalvm-installer/20.x-java-1.8/configure
@@ -1,6 +1,12 @@
 #!/bin/sh
 set -e
 
+architecture=$(uname -i)
+
+if [ "$architecture" != x86_64 ]; then
+    exit 0;
+fi
+
 SOURCES_DIR=/tmp/artifacts
 SCRIPT_DIR=$(dirname "${0}")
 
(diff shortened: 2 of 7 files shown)


antncl:
I already noticed the issue with the Shell Check. I will check it out next week and re-write it, but please let me know if you are fine with the intended purpose of it.

levivic:
hi @AntStephenson, for the origin openj9 image for z/p, we have to set "Native" env to false so we skip the memory-limit.sh script https://github.com/kiegroup/kogito-images/commit/8148c544e5a2837044fba6b7b0856358e6f3269d. Because this script is introduced by the graalvm modules we disabled for z/p image.

So you probably need to exclude that script for p build.

radtriste (approved):
lgtm once shellcheck is solved

radtriste:
> hi @AntStephenson, for the origin openj9 image for z/p, we have to set "Native" env to false so we skip the memory-limit.sh script [8148c54](https://github.com/kiegroup/kogito-images/commit/8148c544e5a2837044fba6b7b0856358e6f3269d). Because this script is introduced by the graalvm modules we disabled for z/p image.
> 
> So you probably need to exclude that script for p build.

@levivic AFAIK we don't have native image on p. So the script should never be called or ?

levivic:
@radtriste you are right, no natives on p/z, which is different than x86. And this needs to be considered when sharing the same override files (builder and jvm runtime) among all three platforms.

spolti (approved):
sounds good to me,
before merge please see the shellsheck issues.

levivic:
Just a correction to my previous comment as per discussion yesterday, seems "NATIVE" evn is not set true on all platforms at this moment (https://github.com/kiegroup/kogito-images/blob/dbd8f53f7e7c5230dd8002ca817f8d4c0d1e98bf/rhpam-kogito-builder-rhel8-overrides.yaml#L27-L29), so it should not affect this PR.

radtriste:
@AntStephenson we would need a rebase of your PR here

antncl:
@spolti if you are happy can I/you merge this? I will cherry pick the squashed commit and create a main branch PR afterwards

spolti:
@AntStephenson thanks for this PR
the changes are good to me, please squash your commits and cherry pick on 1.13.x and main branches.

antncl:
[1.13.x] https://github.com/kiegroup/kogito-images/pull/1149
[main] https://github.com/kiegroup/kogito-images/pull/1148

PR #1197, opened 2022-03-21
[main] [RHPAM-4056] Remove unsupported OpenJ9 images.

antncl:
Remove unsupported OpenJ9 images, they are replaced by multi-arch openjdk images (without s390x arch).

[1.11.x] https://github.com/kiegroup/kogito-images/pull/1195
[1.13.x] https://github.com/kiegroup/kogito-images/pull/1196

@spolti are there any other branches I should raise this for?


spolti:
@AntStephenson is this change correct?
we don't inherit these kind of configuration from openjdk image, we just use a few modules.

antncl:
This PR is just to clean up the repository by removing the previously added (but now unsupported) OpenJ9 images. @levivic wrote that they're now unsupported and won't be built even internally anymore. There was a previous PR a few weeks ago by me to add the multi-arch support for OpenJDK images that was already merged so these can now be removed.

spolti:
we do not use openjdk images, how will it be used?

antncl:
@spolti I thought the existing images do, example here: https://github.com/kiegroup/kogito-images/blob/main/rhpam-kogito-runtime-jvm-rhel8-overrides.yaml#L40 ?

antncl:
I realise we don't use the openjdk images directly (probably what you were saying) but it just enables multi-arch compilation for the existing image and I guess those parent ubi images are available for that arch

spolti:
> @spolti I thought the existing images do, example here: https://github.com/kiegroup/kogito-images/blob/main/rhpam-kogito-runtime-jvm-rhel8-overrides.yaml#L40 ?

the openjdk modules does not uses the openjdk images, this module only installs the openjdk binary.

spolti:
> I realise we don't use the openjdk images directly (probably what you were saying) but it just enables multi-arch compilation for the existing image and I guess those parent ubi images are available for that arch

I don't know how it will be handled on build side, but if that's ok to remove it, we can merge this.

ghatwala:
from previous PR - https://github.com/kiegroup/kogito-images/pull/829/files which was adding openJ9 support, except keeping power support in `content_sets`, rest all seems to be removed which looks good to me. @levivic  - wdyt ?

levivic:
@AntStephenson @ghatwala , removing all semeru modules and the overrides file seems good to me. 
@spolti I guess we are having a bit confusing here, lol. There is no openjdk image involved here, like you said it's an image with openjdk binary installed, so we'd better call it openjdk based images from now on.

spolti:
is this good to merge then?

levivic:
lgtm

PR #1705, opened 2023-10-13
[kie-issues#628] Change JBoss Nexus to Apache Nexus.

baldimir:
Changes references to JBoss Nexus to Apache Nexus.


jstastny-cz (approved):
LGTM.

spolti:
Does the Apache Nexus have the Quarkus dependencies?

ricardozanini:
@baldimir I also share the same concerns if the required libs can be found there. I'm pretty positive that Apache Nexus has a central mirror, so we should be fine.

Regarding the CI:

```
Could not find credentials entry with ID 'kie-ci1-token'
```

Do we need to do anything from our side? It seems a config problem.

jstastny-cz:
Jenkins job failure related to apache/incubator-kie-issues#632 . I'll take that one now, if you can cope without the PR check, feel free to merge as soon as you have necessary reviews

baldimir:
@ricardozanini are these changes fine please? If yes, please approve.

ricardozanini:
@baldimir sure, do we know why CI is red?

Seems like a GH API problem:

```
+ wget --no-check-certificate -qO - GITHUBapi/json?depth=0
```

baldimir:
As far as I understand (please correct me if wrong @jstastny-cz), the Jenkins PR check for kogito-images needs to be rewritten as it is using an unsupported Jenkins plugin.

ricardozanini:
@baldimir how do we check then that this change won't bring bugs to the images?

baldimir:
I see there are some GHA checks, are those working properly? Do you know please? 

For the change itself - we cannot release in Apache, if we use JBoss Nexus. So even with broken Jenkins PR check, we need this change.

ricardozanini:
@baldimir as discussed, let's try to fix the CI first since it does the integration tests we need to make sure everything is fine with these changes. The GHAs checks are more unit/static verifications. I believe we should assure at least the IT before merging this.

jstastny-cz:
This all has hopefully become part of
* apache/incubator-kie-kogito-images#1709

And this one can be closed.

PR #1716, opened 2023-12-12
kie-issues#759: make git clone pattern more robust

jstastny-cz:
Fixes
- apache/incubator-kie-issues#759

Tweak the pattern to replace git clone command only up to (including) the url, leaving possible path argument intact.


jstastny-cz:
I killed the jenkins job, not relevant.

PR #1781, opened 2024-07-19
[10.0.x] NO-ISSUE: set 10.0.x as a tag

jstastny-cz:
Adjusting manage-kogito-version.py script to allow bumping versions to 10.0.x

The value `args.bump_to`, when containing `10.0.x`, might not be valid for all place where applied - because the value is also used as defaults e.g. for `artifacts_version` (maven versions), but script offers sufficient override options.

By allowing 10.0.x and adjusting setup-branch job to actually set such value, we don't need any special precautions when building nightlies (like prefixing a tag or what not). Actually those prefixes `nightly-` were removed some time ago, when we agreed that nightly tag would actually be `10.0.x` for `10.0.x` branches.

### DIFF
--- a/.ci/jenkins/Jenkinsfile.setup-branch
+++ b/.ci/jenkins/Jenkinsfile.setup-branch
@@ -53,7 +53,7 @@ pipeline {
             steps {
                 script {
                     dir(getRepoName()) {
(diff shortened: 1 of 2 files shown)


rodrigonull (approved):
LGTM. Thanks!

tiagobento (approved):

tiagobento:
@jstastny-cz What about the weekly jobs? They need to change too, no?

